PR #12, opened 2018-03-07
Add Smartlock feature

### LINKED ISSUE #2: Support Google SmartLock for password retrieval
Implementatin docs: https://developers.google.com/identity/smartlock-passwords/android/

We have it at VG today so we don't want to regress on this. Could be an optional feature in the SDK. "Worst case" you could just expose the username and password fields in the login/signup form (since they are two forms now you might need to be a bit creative here)

Astach:
- Launch smartlock whatever the chosen flow ( password/passwordless)
- Silently log the user in if credentials are found
- If the user aborts the smartlock login -> fallback to the launched flow
- If there are multiple accounts, the user will be prompt to choose one, if it's the first login the selected flow will be launch and the identifier prefilled.
-if the selected flow is the password one, the use will be asked to save the credentials ( doesn't work for passwordless since it makes no sense to save a temporary validation code)
- if the login fails with smartlock, credentials will be deleted.

Fixes #2


common/src/main/java/com/schibsted/account/common/smartlock/Smartlock.kt
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (c) 2018 Schibsted Products & Technology AS. Licensed under the terms of the MIT license. See LICENSE in the project root.
+ */
+
+package com.schibsted.account.common.smartlock
+
+interface Smartlock {
+    fun requestCredentials()
+    fun saveCredential(username: String, password: String)
```
Nexcius: `saveCredentials`

common/src/main/java/com/schibsted/account/common/smartlock/Smartlock.kt
```diff
@@ -0,0 +1,11 @@
...
+    fun deleteCredential()
```
Nexcius: +s

core/src/main/java/com/schibsted/account/ClientConfiguration.kt
```diff
@@ -7,9 +7,9 @@ package com.schibsted.account
 import android.os.Parcel
 import android.os.Parcelable
 import android.support.annotation.VisibleForTesting
+import com.schibsted.account.common.util.Logger
```
Nexcius: Re-arranging imports should ideally not be done in untouched classes

core/src/main/java/com/schibsted/account/engine/controller/PasswordlessController.kt
```diff
@@ -12,13 +12,13 @@ import com.schibsted.account.engine.input.VerificationCode
...
 import com.schibsted.account.engine.integration.ResultCallback
 import com.schibsted.account.engine.integration.contract.PasswordlessContract
-import com.schibsted.account.engine.operation.AccountStatusOperation
-import com.schibsted.account.engine.operation.AgreementLinksOperation
 import com.schibsted.account.engine.operation.AgreementsCheckOperation
 import com.schibsted.account.engine.operation.MissingFieldsOperation
-import com.schibsted.account.engine.operation.ResendCodeOperation
-import com.schibsted.account.engine.operation.SendValidationCodeOperation
 import com.schibsted.account.engine.operation.VerifyCodeOperation
+import com.schibsted.account.engine.operation.AccountStatusOperation
```
Nexcius: Should not be part of the changeset

example/src/main/java/com/schibsted/account/example/MainActivity.java
```diff
@@ -32,6 +31,8 @@
 
 import java.util.Locale;
 
+import static com.schibsted.account.ui.smartlock.SmartlockImpl.SMARTLOCK_FAILED;
```
Nexcius: Please avoid static imports

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,4 +1,4 @@
-#Mon Oct 30 12:13:42 CET 2017
+#Fri Feb 23 15:42:55 CET 2018
```
Nexcius: Unneeded change

smartlock/build.gradle
```diff
@@ -0,0 +1,86 @@
...
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        }
+    }
+
+}
+
+dependencies {
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
```
Nexcius: Should be removed

smartlock/build.gradle
```diff
@@ -0,0 +1,86 @@
...
+    implementation project(":common")
+    compileOnly project(":core")
```
Nexcius: Why is this compile only?

smartlock/build.gradle
```diff
@@ -0,0 +1,86 @@
...
+                fixDependencies(dependenciesNode, project)
+                addDependency(dependenciesNode, artifactOrg, "account-sdk-android-common", artifactVersion)
+                addArtifactMetadata(asNode(), "Schibsted Account SDK Smartlock Module",
+                        "The Smartlock module for the Schibsted Account SDK", "https://github.com/schibsted/account-sdk-android", pomConfig)
+            }
+        }
+    }
+}
+
+bintray {
```
Nexcius: This will fail without a Bintray package. Please set one up in https://bintray.com/schibsted/Account-SDK-Android
(create an account and send me your username for the org invite)

smartlock/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
```
Nexcius: Should be empty

smartlock/src/main/java/com/schibsted/account/smartlock/SmartlockController.kt
```diff
@@ -0,0 +1,171 @@
...
+    /**
+     * Request Credentials from the Credentials API.
+     */
+    private fun requestCredentials(credentialRequest: CredentialRequest) {
+
+        credentialsClient.request(credentialRequest).addOnCompleteListener(
+                OnCompleteListener<CredentialRequestResponse> { task ->
+                    if (task.isSuccessful) {
+                        retrieveCredential(task.result.credential)
+                        return@OnCompleteListener
```
Nexcius: Simple if .. else will suffice :)

smartlock/src/main/java/com/schibsted/account/smartlock/SmartlockController.kt
```diff
@@ -0,0 +1,171 @@
...
+                    }
+                })
+    }
+
+    private fun retrieveCredential(credential: Parcelable?) {
+        val cred = credential as? Credential
+        cred?.let {
+            val id = Identifier(Identifier.IdentifierType.EMAIL, credential.id)
+            val password = credential.password ?: ""
+            if (password.isEmpty()) {
```
Nexcius: Could this ever be empty?

ui/README.md
```diff
@@ -29,14 +29,15 @@ All flows requires a `UiConfiguration` object to be initialized. This can be cre
 - `locale` the locale to use for the UIs. Defaults to `Locale.getDefault()`
 - `identifierType` which identifier to use for the UIs. `Identifier.IdentifierType.EMAIL` or `Identifier.IdentifierType.SMS`. Defaults to email.
 - `signUpEnabled` option to enable or disable sign-up, only allowing existing users. Defaults to true.
+- `smartlockEnabled` option to enable or disable smartlock, Defaults to true.
```
Nexcius: Do we need this? Could it just inspect the classpath instead?

ui/src/main/java/com/schibsted/account/ui/login/BaseLoginActivity.kt
```diff
@@ -290,7 +371,7 @@ abstract class BaseLoginActivity : AppCompatActivity(), KeyboardManager, Navigat
             LoginScreen.WEB_NEED_HELP_SCREEN -> R.string.schacc_web_need_help_title
             LoginScreen.VERIFICATION_SCREEN -> R.string.schacc_verification_title
             LoginScreen.WEB_TC_SCREEN -> return
-            else -> throw Resources.NotFoundException("Resource not found")
```
Nexcius: `if (SmartlockImpl.isSmartlockAvailable())` should be part of the when matching or outside of it

ui/src/main/java/com/schibsted/account/ui/login/flow/passwordless/PasswordlessActivity.kt
```diff
@@ -47,7 +48,9 @@ class PasswordlessActivity : BaseLoginActivity(), PasswordlessContract {
         }
         identifierType = super.uiConfiguration.identifierType.value
         navigationController = Navigation(this, this)
-        passwordlessController.start(this)
+        if (credentials == null && !isResolving) {
```
Nexcius: These variables should be renamed as this check does not explain much of what it does

ui/src/main/java/com/schibsted/account/ui/smartlock/SmartlockImpl.kt
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    override fun onFailure() {
+        val intent = loginActivity.intent.putExtra(KEY_UI_CONFIGURATION, loginActivity.uiConfiguration.newBuilder().disableSmartlock().build())
+        loginActivity.setResult(SMARTLOCK_FAILED, intent)
+        loginActivity.navigationController.finishNavigation()
+    }
+
+    companion object {
+        const val RC_READ = 3
```
Nexcius: These codes should be documented

ui/src/main/java/com/schibsted/account/ui/login/BaseLoginActivity.kt
```diff
@@ -119,10 +164,16 @@ abstract class BaseLoginActivity : AppCompatActivity(), KeyboardManager, Navigat
...
             BaseLoginActivity.tracker?.flowVariant = TrackingData.FlowVariant.PASSWORD
         }
+    }
 
-        fragmentProvider = FragmentProvider(uiConfiguration)
-
-        followDeepLink(intent.dataString)
+    private fun initializeSmartlock() {
+        val isSmartlockReady = SmartlockImpl.isSmartlockAvailable() && uiConfiguration.smartlockEnabled
+        progressBar.visibility = if (isSmartlockReady) VISIBLE else GONE
```
Nexcius: Move this to if/else block

ui/README.md
Astach: We need this in case of smartlock failure, the flow has to be relaunched, but without smarlock enabled. We can't do this ourself (if we do we have to let the old our activity opened to catch onActivityResult, and we should tell the client which request code we use)

smartlock/src/main/java/com/schibsted/account/smartlock/SmartlockController.kt
Astach: yes, if smart-lock finds a __single__ google account on the device the task will be successful and `Credential` returned, but if the user never saved his credentials with smartlock the password will be `null`, only the identifier can be retrieved. In this case, we prefill identifier only.

ui/src/main/java/com/schibsted/account/ui/UiConfiguration.kt
```diff
@@ -10,56 +10,61 @@ import android.os.Parcel
...
+    val teaserText: String? = null,
+    val signUpNotAllowedErrorMessage: String? = null,
+    val isClosingAllowed: Boolean = true
 ) : Parcelable {
 
     init {
         if (!signUpEnabled && signUpNotAllowedErrorMessage.isNullOrEmpty()) {
             throw IllegalArgumentException("The property signUpNotAllowedErrorMessage must be specified if signUpEnabled is set to false")
         }
+        if (!SmartlockImpl.isSmartlockAvailable()) {
```
Nexcius: Missing check for `smartlockEnabled` before `isSmartlockAvailable`

smartlock/smartlock-rules.pro
```diff
@@ -0,0 +1,3 @@
+-keep com.schibsted.account.smartlock.SmartlockController { public protected *; }
+-keep class com.schibsted.account.ui.smartlock.SmartlockImpl$** { *; }
```
Nexcius: This can be removed

ui/src/main/java/com/schibsted/account/ui/UiConfiguration.kt
```diff
@@ -33,7 +33,7 @@ data class UiConfiguration(
         if (!signUpEnabled && signUpNotAllowedErrorMessage.isNullOrEmpty()) {
             throw IllegalArgumentException("The property signUpNotAllowedErrorMessage must be specified if signUpEnabled is set to false")
         }
-        if (!SmartlockImpl.isSmartlockAvailable()) {
+        if (smartlockEnabled && !SmartlockImpl.isSmartlockAvailable()) {
             throw IllegalAccessException("The smartlock dependency wasn't found")
```
Nexcius: Should be `IllegalStateException`
Please add to the description why this happens. Something like: Smartlock was enabled, but not found on the classpath. Please verify that the smartlock module is included in your build.

PR #16, opened 2018-03-08
Add classloader to Parcel.readStack to avoid using the default one

Astach:
When using a default class loader reading a Parcel object fails.
How to reproduce : 
1- Take any samsung device (of course)
2- launch a passwordless flow
3- go to the verification screen
4 - trigger onPause()/onResume() by clicking on the task manager button
Hurra, app will crash onResume when trying to read the parcel.

### DIFF
--- a/common/src/main/java/com/schibsted/account/common/util/Util.kt
+++ b/common/src/main/java/com/schibsted/account/common/util/Util.kt
@@ -55,9 +55,9 @@ fun String.safeUrl(): String {
     return domain + hiddenParams
 }
 
-fun <T> Parcel.readStack(): Stack<T> {
+fun <T> Parcel.readStack(loader: ClassLoader): Stack<T> {
     val items = mutableListOf<T>()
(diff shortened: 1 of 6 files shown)


example/src/main/java/com/schibsted/account/example/MainActivity.java
```diff
@@ -66,7 +65,9 @@ protected void onCreate(Bundle savedInstanceState) {
         identityReceiver = new IdentityReceiver();
 
         // Start the flow
-        startActivityForResult(passwordIntent, PASSWORD_REQUEST_CODE);
+        if (savedInstanceState == null) {
```
Nexcius: This change is already in #15

Astach: Good catch !

Nexcius:
Probably better to reset `MainActivity` to origin/master and rebase

PR #20, opened 2018-03-09
Add API routes for SPiD actions

Nexcius:


aliak00:
Paths LGTM 👍 I'll leave the kotlin magic to antoine.

core/src/main/java/com/schibsted/account/util/DeepLink.kt
```diff
@@ -74,10 +73,10 @@ sealed class DeepLink {
                 return null
             }
 
-            fun create(redirectUri: URI, identifier: String): String {
-                return URLEncoder.encode("$redirectUri?" +
+            fun create(redirectUri: URI, identifier: String): URI {
```
Astach: the name must be changed `DeepLink.create(..)` strongly suggest that an instance of `DeepLink` class is gonna be returned, I'd prefer something like `createDeepLinkUri` for example

example/src/main/java/com/schibsted/account/example/MainActivity.java
```diff
@@ -40,6 +41,8 @@
     private TextView userState;
     private Button logoutButton;
 
+
```
Astach: :fearful:

PR #31, opened 2018-03-15
Fix issue where parsing profile data would fail

Nexcius:


core/src/main/java/com/schibsted/account/util/LenientAccountsDeserializer.kt
```diff
@@ -0,0 +1,20 @@
+package com.schibsted.account.util
+
+import com.google.gson.JsonDeserializationContext
+import com.google.gson.JsonDeserializer
+import com.google.gson.JsonElement
+import com.google.gson.reflect.TypeToken
+import com.schibsted.account.network.response.ProfileData
+import java.lang.reflect.Type
+
+class LenientAccountsDeserializer : JsonDeserializer<Map<String, ProfileData.Address>> {
```
Astach: Why Lenient ?

PR #45, opened 2018-03-19
Change behavior to set the language of the application dependending on the uiConfiguration

### LINKED ISSUE #43: UI's language should be decided from configuration, not system language

Astach:
Fixes #43 
I do know that this method is deprecated but we can't use the new one. The new one gives us a context to set in `attachBaseContext` unfortunately when this method is called it's too early to create/get the `UIConfiguration`. (I tried it with the following function)

```
    @JvmStatic
    fun getContext(context: Context, locale: Locale): ContextWrapper {
        val resources = context.resources
        var newContext = context
        val conf = resources.configuration
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR2) {
            conf.setLocale(locale)
            newContext = context.createConfigurationContext(conf)
        } else {
            conf.locale = locale
            resources.updateConfiguration(conf, resources.displayMetrics)
        }
        return ContextWrapper(newContext)
    }
```

### DIFF
--- a/example/src/main/java/com/schibsted/account/example/MainActivity.java
+++ b/example/src/main/java/com/schibsted/account/example/MainActivity.java
@@ -54,6 +54,7 @@ public class MainActivity extends AppCompatActivity {
         final UiConfiguration uiConfiguration = UiConfiguration.Builder.fromManifest(getApplicationContext())
                 .enableSignUp()
                 .logo(R.drawable.ic_example_logo)
+                .locale(new Locale("nb", "NO"))
                 .teaserText(getString(R.string.example_teaser_text))
                 .build();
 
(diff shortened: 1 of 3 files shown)


Nexcius:
Not sure I understand why `it's too early to create/get the UIConfiguration` and we need to use the deprecated variant?

Astach:
`attachBaseContext(newBase: Context?)` is called before `onCreate()` meaning we can't get the `Uiconfiguration` (because we're getting the configuration from intent from `onCreate`).
If we want to use the not deprecated methods, which gives us a context to use in `attachBaseContext(newBase: Context?)` we need to have the `Uiconfiguration` to get the locale, and we don't

PR #46, opened 2018-03-20
Add possibility to customize toolbar colors

### LINKED ISSUE #42: Add toolbar icons configuration
Currently we can override toolbar and other SPiD UI texts with color resources. But there is no option to customize colors of icons displayed in toolbar (back arrow, close button) other than overriding these drawables with same names.

Please provide an option to customize colors of them. It will be enough to just add `android:tint` to icons in toolbar of `schacc_mobile_activity_layout` and link them to specific color resource. That way, overriding this color resource would make them unified with the rest of the app.

Astach:
Fixes #42


Astach:
Done, all icons share the same colour

PR #47, opened 2018-03-20
Change default state of keep me logged in option to true

Astach:


example/src/main/java/com/schibsted/account/example/MainActivity.java
```diff
@@ -58,7 +58,7 @@ protected void onCreate(Bundle savedInstanceState) {
                 .build();
 
         // Create the intent for the desired flow
-        final Intent passwordIntent = PasswordActivity.getCallingIntent(this, uiConfiguration);
+        final Intent passwordIntent = PasswordlessActivity.getCallingIntent(this, uiConfiguration);
```
Nexcius: Intentional change?

Astach: Nope good catch !

PR #55, opened 2018-03-23
Fix smartlock dependency

Astach:


smartlock/smartlock-rules.pro
```diff
@@ -1 +1 @@
--keep com.schibsted.account.smartlock.SmartlockController { public protected *; }
+-keep class com.schibsted.account.smartlock.SmartlockController { public protected *; }
```
Nexcius: Why is this kept?

smartlock/src/main/java/com/schibsted/account/smartlock/SmartlockController.kt
```diff
@@ -81,14 +82,12 @@ class SmartlockController(private val activity: AppCompatActivity, private val s
...
         cred?.let {
-            val id = Identifier(Identifier.IdentifierType.EMAIL, credential.id)
             val password = credential.password ?: ""
             if (password.isEmpty()) {
                 smartLockCallback.onHintRetrieved(credential.id)
                 Logger.info(TAG, { "hint was successfully retrieved" })
             } else {
-                val identityCredentials = com.schibsted.account.engine.input.Credentials(id, password, true)
-                smartLockCallback.onCredentialRetrieved(identityCredentials)
+                smartLockCallback.onCredentialRetrieved(credential.id, password, true)
```
Nexcius: Should we consider using a separate data class with an optional password?

ui/ui-rules.pro
```diff
@@ -1 +1,3 @@
 -printusage
+-dontwarn com.schibsted.account.smartlock.SmartlockController
```
Nexcius: This should not be in ui-rules, but rather smartlock-rules

smartlock/smartlock-rules.pro
Astach: To check the path in order to know if Smarlock is available or not, but a keyword was missing

ui/ui-rules.pro
Astach: No, when UI generatedex is running, that checks for classes and can't find SmartlockController in the UI module (because if you don't have the smarlock dependency you'll not find it) and then that crashes. Putting this rule in the `smartlock-rules` will not resolve the problem because the `smartlock-rules`  because the problem comes from the UI module and `smartlock-rules` will not be run if the dependency isn't there anyhow. I had the same thought at the beginning, so I know for sure it doesn't work because I tried it :smile:

smartlock/src/main/java/com/schibsted/account/smartlock/SmartlockController.kt
Astach: That will just move the check elsewhere, I think I prefer to have two separate callbacks, that looks easier to understand. what do you think?

Nexcius: Yeah, that's probably better!

PR #63, opened 2018-03-26
Fix issue where user was not logged in after a successful deeplink signup redirection

### LINKED ISSUE #62: Fails to log the user in after successful deep link redirection
When trying to parse the deeplink there's a check :
` it -> it.host == null && it.path.startsWith("/login") || it.host == "login" }` but in somes cases ( like Aftenposten) the path is "/login" and the host is not null (it's "spid") ` it.host == null ` has to be removed from the check

Astach:
Fixes #62

### DIFF
--- a/core/src/main/java/com/schibsted/account/util/DeepLinkHandler.kt
+++ b/core/src/main/java/com/schibsted/account/util/DeepLinkHandler.kt
@@ -26,7 +26,9 @@ object DeepLinkHandler {
         val uri: URI? = Try { URI.create(dataString) }.getOrElse {
             Logger.info(TAG, { "Unable to parse deep link: ${dataString.safeUrl()}" }, it)
             null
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/schibsted/account/util/DeepLinkHandler.kt
```diff
@@ -26,7 +26,7 @@ object DeepLinkHandler {
         val uri: URI? = Try { URI.create(dataString) }.getOrElse {
             Logger.info(TAG, { "Unable to parse deep link: ${dataString.safeUrl()}" }, it)
             null
-        }?.takeIf { it -> it.host == null && it.path.startsWith("/login") || it.host == "login" }
+        }?.takeIf { it -> it.path.startsWith("/login") || it.host == "login" }
```
Nexcius: We need to be restrictive with this. Add the case instead :1313:

PR #81, opened 2018-04-04
 Change example app to be easier to understand

Astach:


core/src/main/java/com/schibsted/account/Events.kt
```diff
@@ -5,9 +5,9 @@
 package com.schibsted.account
 
 object Events {
-    const val ACTION_USER_LOGIN = "AccountSdkActionUserLogin"
+    internal const val ACTION_USER_LOGIN = "AccountSdkActionUserLogin"
     const val ACTION_USER_LOGOUT = "AccountSdkActionUserLogout"
-    const val ACTION_USER_TOKEN_REFRESH = "AccountSdkActionUserTokenRefresh"
+    internal const val ACTION_USER_TOKEN_REFRESH = "AccountSdkActionUserTokenRefresh"
 
     const val EXTRA_USER = "AccountSdkExtraUser"
```
Nexcius: This can also be internal

core/src/main/java/com/schibsted/account/session/User.kt
```diff
@@ -70,6 +69,7 @@ class User(token: UserToken, val isPersistable: Boolean) : Parcelable {
 
                         override fun onSuccess(result: Unit) {
                             callback?.onSuccess(NoValue)
+                            AccountService.localBroadcastManager?.sendBroadcast(Intent(Events.ACTION_USER_LOGOUT).putExtra(Events.EXTRA_USER_ID, userId))
```
Nexcius: We should send this on error as well

ui/src/main/java/com/schibsted/account/ui/login/BaseLoginActivity.kt
```diff
@@ -107,13 +107,11 @@ abstract class BaseLoginActivity : AppCompatActivity(), KeyboardManager, Navigat
...
 
     internal var smartlock: SmartlockImpl? = null
-    private lateinit var accountService: AccountService
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
-        accountService = AccountService(applicationContext)
-        lifecycle.addObserver(accountService)
+        lifecycle.addObserver(AccountService(applicationContext))
```
Nexcius: We need to keep a reference to this to keep in memory, as the observer list uses weak references

PR #89, opened 2018-04-05
Fix issue where app was crashing when getting Agreements

Astach:
I was doing some unrelated tests when I got : 
```
NetworkError(code=-1, type=network_error, description=Expected a com.google.gson.JsonArray but was com.google.gson.JsonPrimitive, endpoint=https://identity-pre.schibsted.com/api/2/terms?<hidden>)
```
So doing a test on postman I discovered that `https://identity-pre.schibsted.com/api/2/terms` doesn't return `summary` as an array anymore but now as a string. That's a good change of course but it made the application crash, would have been great to be noticed :thinking:


Nexcius:
Lets ensure this is the case in PRO and PRO NORWAY as well

Nexcius:
This is being reverted in SPiD PRE, so the change is not needed.

PR #112, opened 2018-04-11
No longer send identifiers through redirect URIs

### LINKED ISSUE #99: Ensure we don't pass identifiers in redirect URIs
We should have a map of 10 id -> uuidv4 with an expiry of 7 days. Use these to pass through redirect URIs

Nexcius:
Closes #99


core/src/main/java/com/schibsted/account/persistence/LocalSecretsProvider.kt
```diff
@@ -0,0 +1,64 @@
+package com.schibsted.account.persistence
```
Astach: Schibsted license blabla missing

core/src/main/java/com/schibsted/account/persistence/LocalSecretsProvider.kt
```diff
@@ -0,0 +1,64 @@
...
+    private val prefs = appContext.getSharedPreferences(SHARED_PREFS_KEY, Context.MODE_PRIVATE)
+
+    private data class Entry(val time: Long, val key: String, val value: String)
+
+    /**
+     * Retrieves previously stored data from a generated key
+     * @param secretKey The secret key for the persisted data
+     * @return The data previously stored or null if no data is available under that key
+     */
+    fun get(secretKey: String): String? {
```
Astach: secret value should be put  outsdie of the file (because reused in `put` function)
so we would have
```

    private val secrets = prefs.getString(SHARED_PREFS_KEY, "[]")?.let {
        GSON.fromJson<List<Entry>>(it, typeToken<List<Entry>>())
    } ?: listOf()
```
and

`  fun get(secretKey: String): String? = secrets.find { it.key == secretKey }?.value`

core/src/main/java/com/schibsted/account/persistence/LocalSecretsProvider.kt
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    /**
+     * Generates a random key for the value you provide and stores the value. If the value already exists,
+     * it will re-use the previously generated id.
+     * @param value The value to be persisted
+     * @return The randomly generated key for your value
+     */
+    fun put(value: String): String {
+        val secrets = prefs.getString(SHARED_PREFS_KEY, "[]")?.let {
```
Astach: with secret as a class property (see comment above) we could have `        val previousResult = secrets.find { it.value == value }?.key?.let{ return it}` instead

core/src/main/java/com/schibsted/account/util/DeepLink.kt
```diff
@@ -59,13 +59,10 @@ sealed class DeepLink {
 
             operator fun invoke(uri: URI): IdentifierProvided? {
                 if (uri.getQueryParam(DeepLinkHandler.PARAM_ACTION) == Action.IDENTIFIER_PROVIDED.value) {
-                    val result = uri.getQueryParam(PARAM_ID)
-                            ?.let { decodeBase64(it) }
-                            ?.takeIf { Patterns.EMAIL_ADDRESS.matcher(it).matches() }
-                            ?.let { IdentifierProvided(it) }
+                    val result = uri.getQueryParam(PARAM_ID)?.let { IdentifierProvided(it) }
```
Astach: sure encoding isn't needed any longer?

core/src/test/java/com/schibsted/account/persistence/LocalSecretsProviderTest.kt
```diff
@@ -0,0 +1,83 @@
...
+import java.util.UUID
+
+class LocalSecretsProviderTest : StringSpec({
+    val data = mutableMapOf<String, String>()
+
+    val mockEditor: SharedPreferences.Editor = mock {
+        on { putString(any(), any()) }.then {
+            val arg1 = it.getArgument(0) as String
+            val arg2 = it.getArgument(1) as String
+            data.put(arg1, arg2)
```
Astach: could be `data[arg1] = arg2`

core/src/test/java/com/schibsted/account/persistence/LocalSecretsProviderTest.kt
```diff
@@ -0,0 +1,83 @@
...
+
+    }
+
+    "Storing data should return a secret key in the UUID format" {
+        val lsp = LocalSecretsProvider(mockedContext)
+        val res = lsp.put("somedata")
+        UUID.fromString(res)
+    }
+
+    "Storing an existing data value should returnthe same key" {
```
Astach: missing space

core/src/test/java/com/schibsted/account/persistence/LocalSecretsProviderTest.kt
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        on { edit() }.thenReturn(mockEditor)
+    }
+
+    val mockedContext: Context = mock {
+        on { getSharedPreferences(any(), any()) }.thenReturn(mockedSharedPreferences)
+    }
+
+    "Retrieveing data which does not exist should not crash" {
```
Astach: Typo `Retrieving `

core/src/test/java/com/schibsted/account/persistence/LocalSecretsProviderTest.kt
```diff
@@ -0,0 +1,83 @@
...
+        val lsp = LocalSecretsProvider(mockedContext)
+        lsp.get("someKey") shouldBe null
+
+    }
+
+    "Storing data should return a secret key in the UUID format" {
```
Astach: Is the assertion missing there?

core/src/test/java/com/schibsted/account/persistence/LocalSecretsProviderTest.kt
```diff
@@ -0,0 +1,83 @@
...
+        val lsp = LocalSecretsProvider(mockedContext)
+        val myData = "thisIsMydataThereAreManyLikeIt"
+        val res = lsp.put(myData)
+        lsp.get(res) shouldEqual myData
+    }
+
+    "When exceeding the max items, the oldest should be dropped" {
+        val lsp = LocalSecretsProvider(mockedContext, 3)
+
+        val table = table(
```
Astach: :+1:

ui/src/main/java/com/schibsted/account/ui/login/screen/password/PasswordFragment.kt
```diff
@@ -59,8 +61,14 @@ class PasswordFragment : FlowFragment<PasswordContract.Presenter>(), PasswordCon
         mobile_password_button_forgot.setOnClickListener {
             BaseLoginActivity.tracker?.eventEngagement(TrackingData.Engagement.CLICK, TrackingData.UIElement.FORGOT_PASSWORD, TrackingData.Screen.PASSWORD)
 
-            val redirectUri = DeepLink.IdentifierProvided.createDeepLinkUri(uiConf!!.redirectUri, identifier?.identifier
-                    ?: "")
+            val idKey = identifier?.let {
+                with(LocalSecretsProvider(requireContext())) {
```
Astach: `with` not needed here

core/src/main/java/com/schibsted/account/util/DeepLink.kt
Nexcius: Yeah, no special characters in an uuid

core/src/test/java/com/schibsted/account/persistence/LocalSecretsProviderTest.kt
Nexcius: `UUID.fromString(res)` would crash if its not a valid uuid, so that is the test

Astach: Ahhhh

PR #116, opened 2018-04-13
Add support for multiple simultaneous users

### LINKED ISSUE #82: Not supporting multiple users due to ServiceHolder
As we're using the `ServiceHolder` to cache the services, we only support one user at a time. We should consider creating the services from the user object instead

Nexcius:
Fixes #82


core/src/main/java/com/schibsted/account/network/ServiceHolder.kt
```diff
@@ -4,105 +4,24 @@
...
 import com.schibsted.account.network.service.passwordless.PasswordlessService
-import com.schibsted.account.network.service.session.SessionService
-import com.schibsted.account.network.service.user.UserService
-import com.schibsted.account.session.User
 import okhttp3.OkHttpClient
 import java.util.concurrent.TimeUnit
 
 object ServiceHolder {
-    private const val TIMEOUT_MS = 10_000L
+    internal const val TIMEOUT_MS = 10_000L
```
Astach: why?

PR #117, opened 2018-04-13
Add a persist function to the user

Nexcius:


core/src/main/java/com/schibsted/account/session/User.kt
```diff
@@ -145,6 +145,13 @@ class User(token: UserToken, val isPersistable: Boolean) : Parcelable {
         }
     }
 
+    /**
+     * Manually persist a user session so that it can be resumed at a later point.
+     */
+    fun persist(context: Context) {
```
Astach: client request?

Nexcius: Sort of, just want to have that available in the public API

PR #137, opened 2018-04-20
Fixing lint issue about service registration

Astach:
If a client try to register the service this error will show up
```
app/src/main/AndroidManifest.xml:16: Error: Class referenced in the manifest, com.schibsted.account.persistence.UserPersistenceService, was not found in the project or the libraries [MissingRegistered]
          <service android:name="com.schibsted.account.persistence.UserPersistenceService" />
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
I first thought of AIDL but as per https://developer.android.com/guide/components/aidl.html it's not required in our case.

> Note: Using AIDL is necessary only if you allow clients from different applications to access your service for IPC and want to handle multithreading in your service. If you do not need to perform concurrent IPC across different applications, you should create your interface by implementing a Binder or, if you want to perform IPC, but do not need to handle multithreading, implement your interface using a Messenger. Regardless, be sure that you understand Bound Services before implementing an AIDL.

Also this commit allow to requires less configuration from the client.

### DIFF
--- a/core/src/main/AndroidManifest.xml
+++ b/core/src/main/AndroidManifest.xml
@@ -6,6 +6,8 @@
     package="com.schibsted.account">
 
     <uses-permission android:name="android.permission.INTERNET" />
-
+    <application>
+        <service android:name="com.schibsted.account.persistence.UserPersistenceService" />
+    </application>
 
 </manifest>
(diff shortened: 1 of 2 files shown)


core/src/main/AndroidManifest.xml
```diff
@@ -6,6 +6,8 @@
     package="com.schibsted.account">
 
     <uses-permission android:name="android.permission.INTERNET" />
-
+    <application>
```
Nexcius: You tested that this gets merged correctly?

Astach: yep

PR #144, opened 2018-04-24
Update README to specify where to find pulse version

Astach:


README.md
```diff
@@ -16,9 +14,12 @@ dependencies {
     // or
     implementation "com.schibsted.account:account-sdk-android-ui:<VERSION>"
     // (optional, available internally in Schibsted only)
-    implementation "com.schibsted.account:account-sdk-android-pulse:<VERSION>"
+    implementation "com.schibsted.account:account-sdk-android-pulse:<PULSE-VERSION>"
```
Nexcius: Could we name this something else, so that it does not get confused with the pulse library version?

Astach: any idea? I put pulse because dependency is account-sdk-android-pulse, ACCOUNT-SDK-PULSE-VERSION?

PR #177, opened 2018-05-14
Convert verification logic from Java to Kotlin

Astach:


ui/src/main/java/com/schibsted/account/ui/login/screen/verification/VerificationPresenter.kt
```diff
@@ -0,0 +1,102 @@
...
+import com.schibsted.account.ui.ui.ErrorField
+import com.schibsted.account.ui.ui.component.CodeInputView
+
+/**
+ * Following the MVP design pattern this interface represent the implementation of the [VerificationContract.Presenter].
+ * this class executes the code verification business logic and ask for UI updates depending on results.
+ */
+class VerificationPresenter(
+    private val verificationView: VerificationContract.View,
+    provider: InputProvider<VerificationCode>
```
Nexcius: Change this to val, no need to set it in the init function

PR #181, opened 2018-05-16
Add scope parameter to resume user from session code method

Astach:


core/src/main/java/com/schibsted/account/network/service/authentication/OAuthService.kt
```diff
@@ -21,16 +21,19 @@ class OAuthService(@Environment environment: String, okHttpClient: OkHttpClient)
...
      * @param redirectUri The uri the redirect targeted.
+     * @param scopes The scopes requested for the session.
      */
     fun tokenFromAuthCode(
         clientId: String,
         clientSecret: String,
         authCode: String,
-        redirectUri: String
+        redirectUri: String,
+        vararg scopes: String
```
Nexcius: Please use a list for this instead, makes it easier when we add another parameter

PR #183, opened 2018-05-16
Add loading screen while fetching client info

Astach:


Nexcius:
Please finish the activity with an error result

PR #194, opened 2018-05-18
Update example UI when sdk is fetching client info

Astach:


example/src/main/java/com/schibsted/account/example/MainActivity.java
```diff
@@ -168,6 +170,9 @@ public void onError(ClientError error) {
                         Toast.makeText(MainActivity.this, error.getMessage(), Toast.LENGTH_SHORT).show();
                     }
                 });
+            }else {
```
okacat: Missing whitespace

PR #203, opened 2018-05-23
Update the documentation to reflect latest changes

Astach:


ui/README.md
```diff
@@ -23,10 +23,10 @@ The minimal configuration of the SDK UIs are: client name, redirect scheme and r
...
+You can further control the behavior of the UIs bu specifying any of the following attributes. These can be specified in your `AndroidManifest.xml` or by implementing `OptionalConfiguration.UiConfigProvider` in your `Application` class. Using only one of these is recommended, but if you were to use both, then the configuration provider will be resolved before the manifest.
 
 - **Locale:** The locale to use for sending verification email and SMS from Schibsted Account. Defaults: `Locale.getDefault()`.
-- **Sign-up enabled:** Wether or not creation of new accounts should be allowed. Please note that an error message must be specified in order to disable this. Default: false.
+- **Sign-up enabled:** whether or not creation of new accounts should be allowed. Please note that an error message must be specified in order to disable this. Default: false.
```
okacat: Nitpicking, but W is not capitalised

ui/README.md
```diff
@@ -23,10 +23,10 @@ The minimal configuration of the SDK UIs are: client name, redirect scheme and r
...
 - **Sign-up disabled message:** The error message to show when a user attempts to create a new account if it's disabled. No default.
```
okacat: Should this be settable through the builder too or are strings a special case?

Astach: it's doable using the builder in java
```
return new OptionalConfiguration.Builder()
                .locale(new Locale("nb", "NO"))
                .clientLogo(R.drawable.ic_example_logo)
                .signUpEnabled(new OptionalConfiguration.SignUpMode.Disabled("Oskar doesn't want you to sign up!"))
                .build();
```

okacat: Aa right haha 😄 Looks like the default is enabled though, maybe the defaults should say "Default: `true` | `SignUpMode.Enabled`" or something like that, since the builder specific info was removed.

Astach: Good catch!

PR #227, opened 2018-05-30
Fix issue with vector images being displayed incorrectly

Nexcius:


Nexcius:
The issue was that some vector graphics were rendered weirdly. Seemed like some calculations were off.

PR #229, opened 2018-05-30
Possible fix for OutOfMemory issue on EditText

Nexcius:


Nexcius:
I don't think this fix will help actually. What happens is an out of memory error when a user inputs their email. Strange thing is that it only happens using Norwegian locale according to the client

okacat:
Huh, did you get a stack trace?

PR #238, opened 2018-06-05
Fix issue where loading dialog dismissal would cause a crash

Nexcius:


core/src/main/java/com/schibsted/account/persistence/UserPersistence.kt
```diff
@@ -106,6 +108,17 @@ internal class UserPersistence(private val appContext: Context) {
         clearTokenCompat()
     }
 
+    /**
+     * Removes sessions with invalid tokens
+     */
+    private fun cleanInvalidTokens() {
+        val (validSessions, invalidSessions) = sessions.partition { it.token.isValidToken() }
```
okacat: Nice!

PR #250, opened 2018-06-07
Changed refresh token to be optional

### LINKED ISSUE #245: Deserializing token response allows for nulls when it shouldn't

Nexcius:
Fix #245


core/src/main/java/com/schibsted/account/session/User.kt
```diff
@@ -122,15 +122,16 @@ class User(token: UserToken, val isPersistable: Boolean) : Parcelable {
             return false
         }
 
-        if (token.refreshToken.isNullOrEmpty()) {
+        val refreshToken = token.refreshToken
+        if (refreshToken == null || refreshToken.isBlank()) {
```
okacat: Why not use `isNullOrBlank()` as above?

Nexcius: Code code inspection does not see it as a null check :/

PR #263, opened 2018-06-11
Possible fix for KeyNotYetValid exception

Nexcius:


core/src/main/java/com/schibsted/account/persistence/PersistenceEncryption.kt
```diff
@@ -42,9 +41,6 @@ class PersistenceEncryption {
         } catch (e: Exception) {
             Logger.error(TAG, { "Failed to decrypt content" }, e)
             null
-        } catch (e: InvalidKeyException) { // Catches KeyNotYetValidException
```
Astach: `KeyNotYetValidException` extend `InvalidKeyException` but `InvalidKeyException` can still be caught, It's used for 
```
 * This is the exception for invalid Keys (invalid encoding, wrong
 * length, uninitialized, etc).
```
so it shouldn't be removed

PR #266, opened 2018-06-11
Remove age disclaimer in case of login

### LINKED ISSUE #262: Age disclaimer is also shown on login
### Summary
The age disclaimer "You need to be at least 16..." is also showing up on login. This should only be shown on account creation. 

Refer to initial task: https://github.com/schibsted/account-sdk-android/issues/134

### Environment
Identity Release 0.6.1 (released May 25th)
Samsung Note 4

### Steps to reproduce
1. Try to log in (existing user) and on the password page, you will see the disclaimer. It should not be there

Astach:
Fixes #262

### DIFF
--- a/ui/src/main/res/layout/schacc_password_fragment_layout.xml
+++ b/ui/src/main/res/layout/schacc_password_fragment_layout.xml
@@ -12,6 +12,7 @@
 
     <TextView
         android:id="@+id/age_limit_info"
+        android:visibility="gone"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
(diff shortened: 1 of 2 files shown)


ui/src/main/res/layout/schacc_password_fragment_layout.xml
```diff
@@ -12,6 +12,7 @@
 
     <TextView
         android:id="@+id/age_limit_info"
+        android:visibility="gone"
```
Nexcius: Having `tools:visibility="visible"` is very useful for working with the UIs :)

PR #297, opened 2018-06-29
Add cache for agreement status check on resume

### LINKED ISSUE #292: Rate limit checks to agreements endpoint on app launch
Currently doing 20K/second and the backend can't handle this load. This has caused the backend team a lot of alerts and has affected backend performance.

> Joakim Wånggren [9:10 PM]
About the sdk ddos against spid, it’s getting out of control. We’re getting hammered daily, loads of alerts on load warnings, and bad experience for everyone, due to the new sdk and agreement checking
We talked about this, and I hoped for a change
Since it’s a monolith, everything else gets affected: Login fails for blocket, LBC complained about privacy broker failures, and us getting 504s.

It's only happening for Android so maybe good to check what iOS has done and synch

Nexcius:
Fixes #292

### DIFF
--- a/core/src/main/java/com/schibsted/account/persistence/StorageDelegate.kt
+++ b/core/src/main/java/com/schibsted/account/persistence/StorageDelegate.kt
@@ -1,0 +1,22 @@
+/*
+ * Copyright (c) 2018 Schibsted Products & Technology AS. Licensed under the terms of the MIT license. See LICENSE in the project root.
+ */
+
+package com.schibsted.account.persistence
+
+import android.content.SharedPreferences
+import kotlin.reflect.KProperty
+
+class StorageDelegate<T>(
+    private val prefs: SharedPreferences,
+    private val getter: SharedPreferences.(String, T) -> T,
+    private val setter: SharedPreferences.Editor.(String, T) -> SharedPreferences.Editor,
+    val defaultValue: T,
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/schibsted/account/persistence/StorageDelegate.kt
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (c) 2018 Schibsted Products & Technology AS. Licensed under the terms of the MIT license. See LICENSE in the project root.
+ */
+
+package com.schibsted.account.persistence
+
+import android.content.SharedPreferences
+import kotlin.reflect.KProperty
+
+class StorageDelegate<T>(
```
Astach: :+1: :+1:

core/src/main/java/com/schibsted/account/persistence/StorageDelegate.kt
```diff
@@ -0,0 +1,22 @@
...
+    private val prefs: SharedPreferences,
+    private val getter: SharedPreferences.(String, T) -> T,
+    private val setter: SharedPreferences.Editor.(String, T) -> SharedPreferences.Editor,
+    val defaultValue: T,
```
Astach: could be private as well

PR #298, opened 2018-06-29
Add way to switch configuration at runtime

Astach:


README.md
```diff
@@ -51,3 +51,18 @@ The environment can be one of `DEV|PRE|PRO|PRO_NORWAY|<CUSTOM_URL>`.
...
+
+#### How can I witch environment at runtime?
+1. Put all your environments in `schibsted_account.conf` following the pattern below:
+```yaml
+confId: myEnv1
+environment: PRE
+clientId: 58xxxxxxxxxxxxxxxx27
+clientSecret: k8xxxxxxxxxxxxxLm
+
+confId: myEnv2
```
Nexcius: This should be yaml compliant. I recommend keeping the first 4 as default and add other configs under an `additionalConfigs` array

Nexcius: I suggest that we do per file instead, this way we don't need to change the conf parsing, adding complexity

core/src/main/java/com/schibsted/account/ClientConfiguration.kt
```diff
@@ -40,34 +40,48 @@ data class ClientConfiguration(
...
+            this.configId = configurationId
+            currentConfig = loadConf()
         }
 
         @JvmStatic
         fun set(clientConfiguration: ClientConfiguration) {
             currentConfig = clientConfiguration
         }
 
+        private fun loadConf(): ClientConfiguration {
```
Nexcius: This needs to change to the new solution

core/src/main/java/com/schibsted/account/ClientConfiguration.kt
```diff
@@ -40,34 +40,48 @@ data class ClientConfiguration(
...
+            val assetConf = ConfigurationUtils.paramsFromAssets()
+
+            if (assetConf.size > 1 && configId == DEFAULT_KEY) {
+                throw IllegalArgumentException("You have multiple configuration you need to specify the configuration ID")
+            }
+
+            if (assetConf.containsKey(configId)) {
+                return fromParams(assetConf[configId]!!)
+            } else {
+                throw IllegalArgumentException("$configId wasn't found in the configuration file")
```
Nexcius: Should fall back to the default config

PR #304, opened 2018-07-10
Fix an issue where invalid cache causes a crash

Astach:


core/src/main/java/com/schibsted/account/util/DateUtils.kt
```diff
@@ -0,0 +1,34 @@
...
+import java.util.Locale
+import java.util.Calendar
+import java.util.Random
+
+object DateUtils {
+    private const val DATE_PARSING_FORMAT = "yyyy-MM-dd'T'HH:mm:ss"
+
+    fun fromString(stringDate: String): Date? {
+        var date: Date? = null
+        try {
```
Nexcius: You could use `Try {...}` here

PR #307, opened 2018-07-11
Fix issue where redirect_uri was not correctly built

### LINKED ISSUE #306: deeplink doesn't work for newly created client
### Summary
new client's URI format is `<reverse domain of client>.<clientid>:/login` when the SDK builds the redirect URI  it automatically uses `://`. The extra `/` causes a `redirect_uri_mismatch` which breaks the deeplinks.

Astach:
Fixes #306 
Now we don't automatically put `://` to build the redirect URI but we take what's given by the client.
Because `://` was something static in the code whatever we could do to fix the issue would result in a change from the client.
I have chosen the easiest solution for us and the client by just asking the client to add whether `://` or `:/` to their host. Of course, I was first thinking of "why split this value? let's just ask the client to give us the full self-service redirect URI". Unfortunately, we can't do that because we need the scheme to be added in the *static* AndroidManifest, so we can't resolve the scheme from the full self-service URI and inject it in the `android:scheme` field. I also thought of doing `android:scheme="*"` and do a comparison between the link we caught and the redirect URI given by the client but `android:scheme="*"`  is not allowed (and it makes sense).

### DIFF
--- a/ui/README.md
+++ b/ui/README.md
@@ -14,11 +14,14 @@ __Note:__ You should familiarize yourself with the [Core SDK Readme](../core) wh
 ## Configuration
 The minimal configuration of the SDK UIs are: client name, redirect scheme and redirect host. The client name is displayed on the terms and conditions screen as "I accept the terms and conditions for SPID and _yourAppName_", while the two other fields are required for deep linking and can be found in SelfService.
 
+You can find the redirect scheme and redirect host on self service with the following format `MyschemeMyhost` where `Myscheme` is everything before `:`, for example
+with `spid-myclientid://login` the scheme is `spid-myclientid` and the host is `://login`
+
 `strings.xml`
 ```xml
 <string name="schacc_conf_client_name">My client name</string>
 <string name="schacc_conf_redirect_scheme">spid-myclientid</string>
-<string name="schacc_conf_redirect_host">login</string>
+<string name="schacc_conf_redirect_host">://login</string>
 ``` 
 
 
--- a/ui/README.md
+++ b/ui/README.md
@@ -14,8 +14,8 @@ __Note:__ You should familiarize yourself with the [Core SDK Readme](../core) wh
 ## Configuration
 The minimal configuration of the SDK UIs are: client name, redirect scheme and redirect host. The client name is displayed on the terms and conditions screen as "I accept the terms and conditions for SPID and _yourAppName_", while the two other fields are required for deep linking and can be found in SelfService.
 
-You can find the redirect scheme and redirect host on self service with the following format `MyschemeMyhost` where `Myscheme` is everything before `:`, for example
-with `spid-myclientid://login` the scheme is `spid-myclientid` and the host is `://login`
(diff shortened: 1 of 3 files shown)


ui/README.md
```diff
@@ -14,11 +14,14 @@ __Note:__ You should familiarize yourself with the [Core SDK Readme](../core) wh
 ## Configuration
 The minimal configuration of the SDK UIs are: client name, redirect scheme and redirect host. The client name is displayed on the terms and conditions screen as "I accept the terms and conditions for SPID and _yourAppName_", while the two other fields are required for deep linking and can be found in SelfService.
```
Nexcius: This should probably mention that it should include the separator as well

ui/src/main/java/com/schibsted/account/ui/RequiredConfiguration.kt
```diff
@@ -13,13 +14,17 @@ data class RequiredConfiguration(val redirectUri: URI, val clientName: String) :
...
             val redirectScheme = applicationContext.getString(R.string.schacc_conf_redirect_scheme)
-            val redirectHost = applicationContext.getString(R.string.schacc_conf_redirect_host)
+            var redirectHost = applicationContext.getString(R.string.schacc_conf_redirect_host)
 
             if (clientName.isNullOrBlank()) throw IllegalArgumentException("A value for the string <schacc_conf_client_name> must be provided in strings.xml")
             if (redirectScheme.isNullOrBlank()) throw IllegalArgumentException("A value for the string <schacc_conf_redirect_scheme> must be provided in strings.xml")
             if (redirectHost.isNullOrBlank()) throw IllegalArgumentException("A value for the string <schacc_conf_redirect_host> must be provided in strings.xml")
 
-            return RequiredConfiguration(URI.create("$redirectScheme://$redirectHost"), clientName!!)
+            if (!redirectHost.contains(":/")) {
```
Nexcius: I suggest we keep `redirectHost` as a val then move this block to line 17. Something like 
```
with(applicationContext.getString(R.string.schacc_conf_redirect_host)) {
     if (!redirectHost.contains(":/")) { ...
```

ui/src/main/java/com/schibsted/account/ui/RequiredConfiguration.kt
```diff
@@ -14,16 +14,19 @@ data class RequiredConfiguration(val redirectUri: URI, val clientName: String) :
         fun fromResources(applicationContext: Context): RequiredConfiguration {
             val clientName = applicationContext.getString(R.string.schacc_conf_client_name)
             val redirectScheme = applicationContext.getString(R.string.schacc_conf_redirect_scheme)
-            var redirectHost = applicationContext.getString(R.string.schacc_conf_redirect_host)
+            val redirectHost = with(applicationContext.getString(R.string.schacc_conf_redirect_host)) {
+                if (this.contains(":/")) {
```
Nexcius: If you remove the brackets surrounding the ìf` check, you don't need returns

Astach: I have always put brackets with if, especially when one of the branches contains more than one line.
It increases readability, I'm not sure that should block this PR

PR #326, opened 2018-08-12
Decrease complexity of base login activity

Astach:
## Why this PR?
As I was doing a bug fixing, I found out that everything was tied together and that I shouldn't have a so hard time to fix a so easy bug. 
The base activity is responsible for 
- Keyboard Management
- Navigation Management,
- Smartlock management
- Contract implementation and execution
- Fetching client info
- Update the UI depending on one of these things above. 
The problem is that a lot of things interoperate with each other and shouldn't. As an example the result of fetching a clientInfo update the progress bar visibility used in case the smartlock is loading. Also, it's hard to understand the logic behind the code in this activity, I'd say impossible without a good knowledge of the application and its use cases. 
## What was done?
This PR aims to simplify the whole thing and separate concerns,  so it's easier to debug and maintain.
* This PR introduces the use of ViewModel and LiveData (already there in our dependencies, only extensions had to be added). In other words, this PR open the path to the transition from MVP to MVVM, which in a second time will allow us to easily implement data binding.
* Smartlock logic is now in a separate file.
* Keyboard logic is now in a separate file.
* BaseLoginactivty is only responsible for the UI and Navigation.
* SubActivities only implement contracts and start the flow controller(s).
* The view model notifies those activities when something needs to happen.

## What's the gain :muscle: 
- Way easier to understand what's going on, also easier to debug.
- Components are isolated and don't update or interact with non-related objects
- As this was the hardest thing to refactor, the hard work is done, moving to MVVM shouldn't be that hard now.
- UI update will not be done out of the cycle as LiveData is a lifecycle aware component.
- Less boilerplate.

Even if I took care of migrating things without breaking any behaviour I think the next release should be a pre-release, because it's wasn't easy to dig in and refactor this class. Also, it might not be perfect yet, I have chosen a balance between a good cleaning and a not too risky refactoring.


ui/src/main/java/com/schibsted/account/ui/login/flow/password/PasswordActivity.kt
```diff
@@ -4,74 +4,113 @@
...
     private var signUpController: SignUpController? = null
 
-    private val signUpContract = SignupContractImpl(this)
-
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        val savedLoginController = savedInstanceState?.getParcelable<Parcelable>(KEY_LOGIN_CONTROLLER)
-        val savedSignUpController = savedInstanceState?.getParcelable<Parcelable>(KEY_SIGN_UP_CONTROLLER)
+        viewModel.isFlowReady().observe(this, Observer { shouldFetchInfoFirst ->
+            shouldFetchInfoFirst?.let {
```
Nexcius: Could do `if (shouldFetchInfoFirst == true)` not to have to use `let`

Nexcius:
A bit too much for me to catch everything, but LGTM. Wondering if we should have a branch other than master for this

Astach:
Do you think we should do a "beta/Whatchamacallit"  branch from where we can release a beta of the sdk?

PR #328, opened 2018-08-15
Fix issue with behavior on rotation

Astach:
* Use Events Wrapper in order to have catch event only once with live data
* Do not initialize login controller when not needed.

Before this commit - 
* on device rotation user was redirected to the first screen because the ClientResult was emitted again (which cause a navigation to the first screen as intended).
* LoginController was created no matter what, and then the signup flow wasn't working (because handled by the login controller implementation, that means when you were trying to create an account under the hood the app was trying to log you in, and then you got an error saying "Sorry bro, credentials of yours are wrong".

### DIFF
--- a/ui/src/main/java/com/schibsted/account/ui/login/LoginActivityViewModel.kt
+++ b/ui/src/main/java/com/schibsted/account/ui/login/LoginActivityViewModel.kt
@@ -108,7 +108,7 @@ class LoginActivityViewModel(
         if (intentClientInfo == null) {
             fetchClientInfo()
         } else {
-            clientResult.setValue(Event(ClientResult.Success(intentClientInfo)))
+            clientResult.value = Event(ClientResult.Success(intentClientInfo))
         }
     }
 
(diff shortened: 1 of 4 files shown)


ui/src/main/java/com/schibsted/account/ui/login/LoginActivityViewModel.kt
```diff
@@ -105,24 +108,27 @@ class LoginActivityViewModel(
         if (intentClientInfo == null) {
             fetchClientInfo()
         } else {
-            clientResult.value = ClientResult.Success(intentClientInfo)
+            clientResult.setValue(Event(ClientResult.Success(intentClientInfo)))
```
okacat: Why was this changed to use `setValue`?

ui/src/main/java/com/schibsted/account/ui/login/LoginActivityViewModel.kt
```diff
@@ -131,11 +137,8 @@ class LoginActivityViewModel(
         }
     }
 
-    fun isFlowReady(): LiveData<Boolean?> {
-        if (smartlockCredentials.value == null && smartlockReceiver.isSmartlockResolving.value) {
-            return MutableLiveData<Boolean>().apply { value = userFlowType == null }
-        }
-        return MutableLiveData<Boolean>().apply { value = null }
+    fun isFlowReady(): Boolean {
```
okacat: Could be shortened by using `=`

ui/src/main/java/com/schibsted/account/ui/login/LoginActivityViewModel.kt
Astach: Ahah it's just that I forgot to change it. During my refactoring, I reach a state where `ClientResult` was not a `MutableLiveData` anymore but another object with a `setValue` method :sweat:

PR #333, opened 2018-08-27
Do not rely on the App object to get the ui conf

### LINKED ISSUE #329: Since SDK 1.0.1 Locale specified in the code is not respected
### Summary
When I open login screen I can see default language from device instead the one I have specified in the configuration, in the code. 
```
override fun getUiConfig() = OptionalConfiguration.Builder()
            .signUpEnabled(OptionalConfiguration.SignUpMode.Enabled)
            .clientLogo(R.drawable.app_main_logo)
            .locale(configuration.spidLocale)
            .build()
//from our config
override val spidLocale = Locale("nb", "NO")
```

### Environment
- SDK Version: since 1.0.1
- SDK Module: UI I guess
- Android Version: any
- Device Info: any

### Details
I assume that this was broken with this change
`1.0.1`
`Fixed issue where language wasn't set according to the manifest (#293)`

### Steps to reproduce
1. Configure Norwegian in the SDK using code
2. Run login screen on device with Swedish or English

Astach:
Fixes #329 
***What was the problem***
The locale was always retrieved from the manifest even if overridden programmatically. That was due to the call of `OptionalConfiguration.fromManifest` instead of `OptionalConfiguration.revolve`. The last method would apply the configuration correctly but unfortunately we can call it when we need it: We need it in `attachBaseContext` but this `resolve`  method needs an `Application` object which is not created yet.
***What are the changes***
*  The `Application` object is not required any longer to resolve the optional configurations
*  All optional configurations are now defined in one place cf. `AccountUi`
*  The sdk first tries to resolve the locale from the manifest and then tries to resolve it from the runtime conf

### DIFF
--- a/ui/src/main/java/com/schibsted/account/ui/AccountUi.kt
+++ b/ui/src/main/java/com/schibsted/account/ui/AccountUi.kt
@@ -9,6 +9,7 @@ import android.content.Context
 import android.content.Intent
 import android.os.Bundle
 import android.os.Parcelable
+import android.support.annotation.DrawableRes
 import com.schibsted.account.engine.integration.ResultCallback
 import com.schibsted.account.engine.operation.ClientInfoOperation
 import com.schibsted.account.network.OIDCScope
@@ -18,6 +19,7 @@ import com.schibsted.account.ui.login.flow.passwordless.PasswordlessActivity
 import com.schibsted.account.ui.smartlock.SmartlockController
 import com.schibsted.account.ui.smartlock.SmartlockMode
 import kotlinx.android.parcel.Parcelize
+import java.util.Locale
 
 object AccountUi {
     const val KEY_PARAMS = "SCHACC_PARAMS"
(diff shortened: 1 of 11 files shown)


ui/src/main/java/com/schibsted/account/ui/AccountUi.kt
```diff
@@ -42,11 +44,16 @@ object AccountUi {
      *
      */
     @Parcelize
-    data class Params(
+    data class Params internal constructor(
```
Nexcius: This does not need to be internal, for Kotlin developers this makes sense

ui/src/main/java/com/schibsted/account/ui/AccountUi.kt
```diff
@@ -42,11 +44,16 @@ object AccountUi {
...
         val teaserText: String? = null,
         val preFilledIdentifier: String? = null,
         val smartLockMode: SmartlockMode = SmartlockMode.DISABLED,
+        val locale: Locale = Locale.getDefault(),
```
Nexcius: Ensure it's clear that these overrides the manifest values. Either by naming: `overrideLocale` or it might be enough to add to to the function's docs

PR #340, opened 2018-08-30
Remove the default values for InternalUiConfiguration

Astach:
Having default values is error prone. For example when we add a new configuration option in the `InternalUiConfiguration` constructor with a default value but we don't pass the configuration when we construct the object inside `resolve` the compiler will not complain as it's gonna use the default value. So basically it means it's easy to forget to apply the correct configuration when setting it up.

### DIFF
--- a/ui/src/main/java/com/schibsted/account/ui/InternalUiConfiguration.kt
+++ b/ui/src/main/java/com/schibsted/account/ui/InternalUiConfiguration.kt
@@ -15,14 +15,14 @@ import java.util.Locale
 data class InternalUiConfiguration(
     val clientName: String,
     val redirectUri: URI,
-    val locale: Locale = AccountUi.Params.DEFAULT_LOCALE,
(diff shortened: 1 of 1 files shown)


Nexcius:
The problem with this is that we cannot simply override the fields we are interested in and must provide a full configuration. What is the argument for having to provide the full config?

Astach:
The full configuration is always provided in the `resolve` method now,  where the configuraion fields are taken wheter from the manifest, from the dymanic conf or from the default values ( so the default values is now used in the `resolve` method). Removing the default fields values from the constructor force the dev to pass the resolved fields to the constructor, that way we make sure that the right configuration will be used (because otherwise you can resolve a field but forget to pass it to the constructor and in that case the default configuration for this field will always be used no matter what). <- maybe it's better explained in this PR description 😅 . I made the mistake, so I thought it would be nice to fix that and eliminate an occasion to create a bug or at least an occasion to waste time.

zamzterz:
Closing due to inactivity.

Issue #347, opened 2018-09-27
validation not applied on email field


Astach:
### Summary
For email the input is not validate using `EmailValidationRule`
### Environment
- SDK Version: 1.0.1 +
- SDK Module:  UI

### Details
the validation rule should be added in the following section (in `EmailIdentificationFragment`)
```
  inputFieldView = SingleFieldView.create(context!!) {
            isCancelable { true }
            inputType { InputType.TYPE_TEXT_VARIATION_WEB_EMAIL_ADDRESS }
            ime { EditorInfo.IME_ACTION_DONE }
            error { getString(R.string.schacc_email_identification_error) }
            title { getString(R.string.schacc_email_label) }
            hint { getString(R.string.schacc_required_field_email) }
        }
```

### Steps to reproduce
Fill the email field with a not valid email like fix@bug,com

Nilzor:
I'm getting a  screen similar to this when entering wrong email. Is this a consequence of validation not implemented?`

![image](https://user-images.githubusercontent.com/990654/59757094-48bd8480-928b-11e9-9b15-a2abaf5ed2b3.png)

As you can see, this screen is from the register flow - reported by a user who claims he's online. Could it be a similar issue also for the register form? Should I open a separate issue for that?

As you can see

PR #359, opened 2018-11-05
If applied this commit will make sure email are validated.

sa-mao:
Refere to #347 for more details.


ui/src/main/java/com/schibsted/account/ui/login/screen/identification/ui/EmailIdentificationFragment.kt
```diff
@@ -38,6 +38,7 @@ class EmailIdentificationFragment : AbstractIdentificationFragment(), Identifica
             error { getString(R.string.schacc_email_identification_error) }
             title { getString(R.string.schacc_email_label) }
             hint { getString(R.string.schacc_required_field_email) }
+            validationRule { EmailValidationRule }
```
Nexcius: Consider only using this when creating an account. These rules tend to change, so previous valid emails might now be invalid

sa-mao:
I will close this, since we don't currently have a separate sign up step, and we couldn't find a simple way of handling that without confusing the user.

Issue #353, opened 2018-10-09
Focus on username and password box upon enter


Nilzor:
The user should not have to tap the username or password field when you enter those two activities initially. They are not focused by default. They should be.

Astach:
Which version of the SDK?

Nilzor:
`1.5.0`

It also applies to the username form so I'll change the title of the request to that. 

Any chance of having this looked at @idanor ? It sounds like something that should be trivial to fix

idanor:
Hi! Could you explain the problem a bit more @Nilzor ? Difficult to grab the essence of your need by what is described in this thread :)

Nilzor:
Sure I'll try.

## Current flow
- User clicks "log in" in our app
- The Schibsted Account login form is displayed
**- User taps the username field**
- User enters name
- User clicks Next 
- The Schibsted Account password form is displayed
**- User taps the password field**
- User enters password
- User clicks Next
- Schibsted Account flow completed

I want the actions in bold not to be neccessary

idanor:
Thanks, Frode! 

From an Identity product point of view, I have no objections to taking away these steps as long as that works technically. I agree that it would create a better user experience.

sa-mao:
Hi,

We actually already have a [PR](https://github.com/schibsted/account-sdk-android/pull/366) for this, but the previous maintainers advised against implementing it in that way because it might cause inconsistent behaviors across devices. But we would love to get your inputs on the PR.

Nilzor:
Cool, didn't know that. I have now commented in that PR

Mvh
Frode Nilsen
Product Manager for Android apps, CNP - Schibsted Media

Nilzor:
Solved in https://github.com/schibsted/account-sdk-android/pull/407

PR #366, opened 2018-11-13
Make sure we focus on the email and password fields

sa-mao:
This PR will make sure that when you start the login flow, the focus is always on the email and the password field, and you can always immediately start typing.

### DIFF
--- a/ui/src/main/java/com/schibsted/account/ui/login/screen/password/PasswordFragment.kt
+++ b/ui/src/main/java/com/schibsted/account/ui/login/screen/password/PasswordFragment.kt
@@ -85,9 +85,7 @@ class PasswordFragment : FlowFragment<PasswordContract.Presenter>(), PasswordCon
         password_input_view.inputField.requestFocus()
         val imm =  activity!!.applicationContext.getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
         imm.showSoftInput(password_input_view.inputField, InputMethodManager.SHOW_IMPLICIT)
-
-
-
+        
         password_input_view.setImeAction(EditorInfo.IME_ACTION_NEXT) { _, actionId, _ ->
             if (actionId == EditorInfo.IME_ACTION_NEXT) {
                 signUser()
--- a/ui/src/main/java/com/schibsted/account/ui/login/screen/password/PasswordFragment.kt
+++ b/ui/src/main/java/com/schibsted/account/ui/login/screen/password/PasswordFragment.kt
@@ -83,9 +83,9 @@ class PasswordFragment : FlowFragment<PasswordContract.Presenter>(), PasswordCon
         primaryActionView.setOnClickListener { signUser() }
 
         password_input_view.inputField.requestFocus()
-        val imm =  activity!!.applicationContext.getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
+        val imm = activity!!.applicationContext.getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
         imm.showSoftInput(password_input_view.inputField, InputMethodManager.SHOW_IMPLICIT)
-        
+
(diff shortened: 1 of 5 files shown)


ui/src/main/java/com/schibsted/account/ui/login/screen/password/PasswordFragment.kt
```diff
@@ -80,6 +82,10 @@ class PasswordFragment : FlowFragment<PasswordContract.Presenter>(), PasswordCon
         password_input_view.validationRule = (if (isUserAvailable) PasswordValidationRule else BasicValidationRule)
         primaryActionView.setOnClickListener { signUser() }
 
+        password_input_view.inputField.requestFocus()
+        val imm = activity!!.applicationContext.getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
```
Nexcius: Using the IMM like this is usually not a good idea. The Android system actually handles this for you, based on the layout. We used to do this but decided to take it out and let the OS handle it instead. If you want to let this happen automatically, make sure the elements are the first ones (or specify order) in the layout

Nexcius: Also, just calling `password_input_view.inputField.requestFocus()` might be enough

Nilzor:
Hi I got here from https://github.com/schibsted/account-sdk-android/issues/353

From what I read here, @Nexcius is not opposed to solving this, just the way it's solved in the PR. He offers two alternatives:

1. Make sure the username and password fields are the first editable fields in the UI XML list. I suspect they're not
2. calling `password_input_view.inputField.requestFocus()` (and the corresponding for the username field)

ui/src/main/java/com/schibsted/account/ui/login/screen/password/PasswordFragment.kt
Nilzor: I've investigated a bit and just calling `password_input_view.inputField.requestFocus()` is not enough. Not here, neither in a hello world app with only one EditText (just tested).

`android:windowSoftInputMode="stateVisible"` in the manifest doesn't really help us a lot in this case, since we have two fragments with input fields under one Activity. That's why you do the `imm` call in PasswordFragment I suppose, but you forgot to call `requestFocus` in `EmailIdentificationFragment`

I'm not too conserned with calling `imm`. There are a lot of stack overflow threads advocating this solution. There are no other viable options for our use case. The OS does not handle this case. 

However I do want us to test this more and test the UX pattern for different cases. For instance I haven't tested it together with SmartLock enabled. Will test and get back.

zamzterz: Thanks for the follow-up @Nilzor! I did some initial testing on this upon your first comment, and basically came to the same conclusion as you (there's no better way than directly calling the 
`InputMethodManager`).

If you could help out testing this with SmartLock it would be much appreciated, and then hopefully we could land this 👍

Nilzor:
I will get back to this issue after my vacation (mid-august). I want to test it thoroughly on multiple devices with and without smartlock so that we don't make it worse in any cases.

zamzterz:
@Nilzor Any update on the testing of this 🙂

Nilzor:
I'll get back to you by the end of the week

Nilzor:
There are three small issues with this PR: 
- Functionality issue mentioned [here](https://github.com/schibsted/account-sdk-android/pull/366#discussion_r289987255)
- After SmartLock enters username, keyboard pops unnecessarily up (you don't usually type anything when smartlock prefills.)
- Keyboard stays open if the use clicks HOME when in the middle of typing

I'll adress these issues in a separate pull request - I did some refactoring of the code which made it easier just to start over. Will post a new PR soon. Advice to close this

Nilzor:
See #406

zamzterz:
Closed in favor of #406.

PR #369, opened 2018-11-22
Remove usage of codename from release instructions.

zamzterz:


sa-mao:
Maybe we should keep the codename?

PR #375, opened 2019-02-27
Reset keypair if expired

sorbris:
The keypair used for encrypting and decrypting the aes key has a
lifetime of one year. If it expired it was not replaced which caused
session persistance to fail. With this commit the key will be recreated
if for some reason encryption or decryption fails.


sorbris:
Hello @Astach,  
Just some more context around why we are making this change:
  We are about one year out from implementing the Account sdk in the app. Last week we suddenly got a lot of angry users writing reviews and contacting us to say that they had to log in again every time they started the app. This lead to a 14% drop in login rate in a little less than a week.  

Regarding your comments: 
1. In the case of the key being renewed because it is getting close to expiry that is done. See updateKeyExpiry in SessionStorageDelegate for more details. If the key is renewed because the KeyPair has expired that is not possible since we can’t decrypt the stored data with that key anymore.
2. yes
3. The exceptions are guaranteed to happen after a year and is currently not handled but just silently swallowed. They are now propagated so that they can be handled in SessionStorageDelegate. But we will double check to make sure that we catch all of them inside the sdk. 
4. That might be what was intended but that is not what happens. After a year the key expires and the issue is silently swallowed which causes the sdk to fail to persist sessions from that point on since the keypair is expired and never recreated. 
5. As far as I know there is no limitation but we can check with legal.

core/src/main/java/com/schibsted/account/persistence/EncryptionKeyProvider.kt
```diff
@@ -147,6 +183,7 @@ class EncryptionKeyProvider(private val appContext: Context) {
             apply()
         }
         return KeyPair(publicKey, privateKey)
+            .also { persistTimestamp(-1L) }
```
guzno: Just a minor thing, but change this to `NO_EXPIRY`.

core/src/main/java/com/schibsted/account/persistence/EncryptionKeyProvider.kt
```diff
@@ -191,6 +228,10 @@ class EncryptionKeyProvider(private val appContext: Context) {
         private const val SHARED_PREFERENCES_NAME = "IDENTITY_KEYSTORE"
         private const val SHARED_PREFERENCES_PRIVATE_KEY = "IDENTITY_PR_KEY_PAIR"
         private const val SHARED_PREFERENCES_PUBLIC_KEY = "IDENTITY_PU_KEY_PAIR"
+        @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+        internal const val SHARED_PREFERENCES_KEYS_VALID_UNTIL = "KEYS_VALID_UNTIL"
```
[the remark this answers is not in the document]
sorbris: Good point

core/src/main/java/com/schibsted/account/persistence/PersistenceEncryption.kt
```diff
@@ -22,25 +25,33 @@ class PersistenceEncryption {
...
         val cipher = Cipher.getInstance(RSA_TRANSFORM)
         return try {
             cipher.init(Cipher.DECRYPT_MODE, privateRsaKey)
-            return cipher.doFinal(subjectToDecrypt)
+            cipher.doFinal(subjectToDecrypt)
         } catch (e: Exception) {
             Logger.error(TAG, "Failed to decrypt content", e)
-            null
+            if (e is InvalidKeyException || e is IllegalBlockSizeException || e is BadPaddingException) {
+                throw e
```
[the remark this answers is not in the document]
sorbris: I agree

core/src/main/java/com/schibsted/account/persistence/SessionStorageDelegate.kt
```diff
@@ -137,11 +173,23 @@ internal class SessionStorageDelegate(
      * @return the [SecretKey] if the decryption was successful, null otherwise
      */
     private fun getOriginalAesKey(encryptedAesKey: ByteArray): SecretKey? {
-        val aesEncodedKey = encryption.rsaDecrypt(encryptedAesKey, rsaKeyPair.private)
-        if (aesEncodedKey != null && aesEncodedKey.isNotEmpty()) {
-            return SecretKeySpec(aesEncodedKey, 0, aesEncodedKey.size, AES_ALG)
+        return try {
+            getOriginalAesKey(encryptedAesKey, rsaKeyPair.private)
+        } catch (e: Exception) {
```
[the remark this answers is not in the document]
sorbris: I agree

sorbris:
True it does remove the persisted data in `updateKeyExpiry` but it also writes it down again with the new key with writeDataToPersistence. 
The idea is that it should automagically keep the user logged in indefinitely as long as the application is in use. If the app is not in use for the three months leading up to the expiry then you most likely will be logged out.

core/src/main/java/com/schibsted/account/persistence/SessionStorageDelegate.kt
```diff
@@ -24,7 +26,8 @@ internal class SessionStorageDelegate(
     private val preferenceFilename: String,
     private val preferenceKey: String,
     private val encryption: PersistenceEncryption = PersistenceEncryption(),
-    private val rsaKeyPair: KeyPair = EncryptionKeyProvider(appContext).keyPair
+    private val encryptionKeyProvider: EncryptionKeyProvider = EncryptionKeyProvider(appContext),
+    private var rsaKeyPair: KeyPair = encryptionKeyProvider.keyPair
```
zamzterz: Keeping `encryptionKeyProvider.keyPair` as an instance variable in `rsaKeyPair` requires it to be kept synced upon key refresh (which is currently correctly implemented in `refreshKeyPair`). But I'm thinking it might be simpler/better to use `encryptionKeyProvider.keyPair` directly instead which is guaranteed to always be up-to-date.

core/src/main/java/com/schibsted/account/persistence/EncryptionKeyProvider.kt
```diff
@@ -69,6 +83,13 @@ class EncryptionKeyProvider(private val appContext: Context) {
         }
     }
 
+    @SuppressLint("NewApi")
+    internal fun refreshKeyPair(): KeyPair {
```
zamzterz: As suggested below, I think there's no need to keep a copy of the key pair in `SessionStorageDelegate.kt` so this shouldn't return it.

core/src/main/java/com/schibsted/account/persistence/SessionStorageDelegate.kt
sorbris: I agree, I'll fix it

zamzterz:
Superseded by #378 to allow Travis to run the tests.

PR #376, opened 2019-02-27
Initiate SmartLockController on result

sorbris:
If a SmartLockController hasn't been created when handling a result from
SmartLock, one will be created.


sa-mao:
LGTM

sa-mao:
@sorbris I will close this one and replace it by #377 to allow CI execution.

PR #377, opened 2019-02-27
Initiate SmartLockController on result

sa-mao:
If a SmartLockController hasn't been created when handling a result from
SmartLock, one will be created.

Supersedes #376


zamzterz:
Code changes LGTM 👍 .
@sorbris Do you have any instructions for how to reproduce the initial issue (where `smartlockController` is uninitialised ) to verify the changes work as expected?

sorbris:
@zamzterz For us it appeared if the user had multiple viable smartlock entries so the picker was displayed

zamzterz:
@sorbris While testing I found this was an even more general problem: the `SmartLockController` was uninitiated in other cases as well. So I've made #379 to fix it once and for all™️.

Thanks for reporting the issue 👍

PR #381, opened 2019-03-28
Create a device fingerprint after each login

sa-mao:
Create a device fingerprint after each successful login.

### DIFF
--- a/core/src/main/java/com/schibsted/account/AccountService.kt
+++ b/core/src/main/java/com/schibsted/account/AccountService.kt
@@ -33,7 +33,7 @@ class AccountService @JvmOverloads constructor(
                 .getPackageInfo(appContext.packageName, PackageManager.GET_META_DATA)
                 .versionCode
                 .toString()
-        AccountService.androidId =  Settings.Secure.ANDROID_ID
+        AccountService.androidId = Settings.Secure.ANDROID_ID
     }
 
     @OnLifecycleEvent(Lifecycle.Event.ON_START)
--- a/core/src/main/java/com/schibsted/account/AccountService.kt
+++ b/core/src/main/java/com/schibsted/account/AccountService.kt
@@ -9,6 +9,8 @@ import android.arch.lifecycle.LifecycleObserver
 import android.arch.lifecycle.OnLifecycleEvent
 import android.content.Context
 import android.content.Intent
+import android.content.pm.PackageManager
+import android.provider.Settings
 import android.support.annotation.VisibleForTesting
 import android.support.v4.content.LocalBroadcastManager
 import com.schibsted.account.common.util.Logger
@@ -26,6 +28,16 @@ class AccountService @JvmOverloads constructor(
     init {
         AccountService.localBroadcastManager = localBroadcastManager
         AccountService.packageName = appContext.packageName
+        AccountService.packageVersion = appContext
+                .packageManager
+                .getPackageInfo(appContext.packageName, PackageManager.GET_META_DATA)
+                .versionCode
+                .toString()
+        AccountService.androidId =  Settings.Secure
+                .getString(
+                        appContext.contentResolver,
+                        Settings.Secure.ANDROID_ID
+                )
     }
 
     @OnLifecycleEvent(Lifecycle.Event.ON_START)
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/schibsted/account/AccountService.kt
```diff
@@ -26,6 +28,16 @@ class AccountService @JvmOverloads constructor(
...
         AccountService.packageName = appContext.packageName
+        AccountService.packageVersion = appContext
+                .packageManager
+                .getPackageInfo(appContext.packageName, PackageManager.GET_META_DATA)
+                .versionCode
+                .toString()
+        AccountService.androidId =  Settings.Secure
+                .getString(
+                        appContext.contentResolver,
+                        Settings.Secure.ANDROID_ID
```
zamzterz: According to [the docs](https://developer.android.com/reference/android/provider/Settings.Secure.html#ANDROID_ID) this value is "unique to each combination of app-signing key, user, and device [on Android 8.0 and higher]", meaning it will be different per app. Is there some other identifier we can use (on all Android versions we support), that is scoped only by user and device (and not per app)?

core/src/main/java/com/schibsted/account/engine/controller/LoginController.kt
```diff
@@ -55,6 +56,7 @@ class LoginController @JvmOverloads constructor(
             }) { it ->
                 val user = User(it, credentials.keepLoggedIn)
                 this.currentUserId = user.userId
+                user.device.createFingerprint()
```
zamzterz: Could this be tested somehow in `LoginControllerTest`? 🤔

core/src/main/java/com/schibsted/account/session/Device.kt
```diff
@@ -0,0 +1,39 @@
...
+import android.provider.Settings
+import com.schibsted.account.ClientConfiguration
+import com.schibsted.account.engine.integration.ResultCallback
+import com.schibsted.account.model.NoValue
+import com.schibsted.account.model.error.ClientError
+import com.schibsted.account.network.NetworkCallback
+import com.schibsted.account.network.response.DeviceFingerprint
+import com.schibsted.account.network.service.user.UserService
+
+class Device(private val applicationName: String, private val applicationVersion: String, private val androidId: String, private val user: User, private val userService: UserService = UserService(ClientConfiguration.get().environment, user.authClient)) {
```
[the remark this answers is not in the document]
sa-mao: What if we want to introduce a connected devices view in the future? a device will have more to it than just a fingerprint.

core/src/main/java/com/schibsted/account/network/response/DeviceFingerprint.kt
```diff
@@ -0,0 +1,6 @@
+/*
+ * Copyright (c) 2018 Schibsted Products & Technology AS. Licensed under the terms of the MIT license. See LICENSE in the project root.
```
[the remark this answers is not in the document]
sa-mao: Copy paste :D

core/src/main/java/com/schibsted/account/session/User.kt
```diff
@@ -54,6 +54,8 @@ class User(token: UserToken, val isPersistable: Boolean) : Parcelable {
 
     val profile = Profile(this)
 
+    val device = Device(AccountService.packageName, AccountService.packageVersion, AccountService.androidId, this)
```
zamzterz: I don't think there's a need for this to be called outside the SDK? If so, it should be documented or if not it should be `internal`.

core/src/main/java/com/schibsted/account/AccountService.kt
sa-mao: We can use [Identifiers](https://developer.android.com/training/articles/user-data-ids#best-practices-android-identifiers) as an alternative but they are not recommended and there are some other considerations to them (like permissions etc)

core/src/main/java/com/schibsted/account/session/Device.kt
```diff
@@ -1,18 +1,18 @@
...
+ * Copyright (c) 2019 Schibsted Products & Technology AS. Licensed under the terms of the MIT license. See LICENSE in the project root.
+ */
 package com.schibsted.account.session
 
 
-import android.content.Context
-import android.content.pm.PackageManager
-import android.provider.Settings
 import com.schibsted.account.ClientConfiguration
+import com.schibsted.account.common.util.Logger
```
zamzterz: Unused import (but could be nice with some logging here 🙂 ).

core/src/main/java/com/schibsted/account/AccountService.kt
zamzterz: Ok, then this is probably the best we can do for now - let's use it!

PR #387, opened 2019-05-29
Expose Auth Interceptor outside

schibsted-rafal-adamek:
This lets consumers of the SDK to create interceptor on their own OkHttpClient instance instead of relying on `user.bind()` method.

This is mendatory for our project to add own interceptors so we can apply Varnish ID to authenticated requests.


zamzterz:
#389 for Travis sake.

Issue #388, opened 2019-06-03
Retreiving stored user credentials fails 100% time on Huawei P20 device


Nilzor:
### Summary
Huawei P20 Pro: Logging in succeeds, but killing the app and restarting it causes the user to be logged out despite of having checked "remember me on this devices".

The error is not reproducable on Samsung S8

### Environment
- SDK Version: 1.6.0
- SDK Module: core/example
- Android Version: 8.1.0
- Device Info: Huawei P20 Pro

### Steps to reproduce
First time install of example app, current master branch. 
- Open example app
- Successfully log in on test environment and check "remember me on this device"
- Kill app
- Restart app.
- *expected*: Fetches credentials from local storage and relogin the user
- *actual* Fails with error message in logcat:

```

2019-06-03 13:20:56.604 5795-5795/com.schibsted.account.example E/SCHACC-PersistenceEncryption: Failed to decrypt content
    android.security.keystore.KeyNotYetValidException: Key not yet valid
        at android.security.KeyStore.getInvalidKeyException(KeyStore.java:727)
        at android.security.KeyStore.getInvalidKeyException(KeyStore.java:793)
        at android.security.keystore.KeyStoreCryptoOperationUtils.getInvalidKeyExceptionForInit(KeyStoreCryptoOperationUtils.java:54)
        at android.security.keystore.KeyStoreCryptoOperationUtils.getExceptionForCipherInit(KeyStoreCryptoOperationUtils.java:89)
        at android.security.keystore.AndroidKeyStoreCipherSpiBase.ensureKeystoreOperationInitialized(AndroidKeyStoreCipherSpiBase.java:265)
        at android.security.keystore.AndroidKeyStoreCipherSpiBase.engineInit(AndroidKeyStoreCipherSpiBase.java:109)
        at javax.crypto.Cipher.tryTransformWithProvider(Cipher.java:2668)
        at javax.crypto.Cipher.tryCombinations(Cipher.java:2575)
        at javax.crypto.Cipher$SpiAndProviderUpdater.updateAndGetSpiAndProvider(Cipher.java:2480)
        at javax.crypto.Cipher.chooseProvider(Cipher.java:567)
        at javax.crypto.Cipher.init(Cipher.java:831)
        at javax.crypto.Cipher.init(Cipher.java:772)
        at com.schibsted.account.persistence.PersistenceEncryption.rsaDecrypt(PersistenceEncryption.kt:45)
[...]

PR #392, opened 2019-07-01
Fix for Retreiving stored user credentials fails 100% time on Huawei P20 device (see #388)

sa-mao:
Use a start date in the past for the encryption key to avoid timezone issue with huawei p20.
See this [thread](https://github.com/iamMehedi/Secured-Preference-Store/issues/15) for more details.

This should fix https://github.com/schibsted/account-sdk-android/issues/388


sa-mao:
Yes that could do as well, It's Mostly to put enough time zones between :D

PR #395, opened 2019-07-05
Added resetting of services to  ClientConfiguration set method.

JakubMosakowski:
If applied, this commit will allow replacing environment in runtime using method "set".

### DIFF
--- a/core/src/main/java/com/schibsted/account/ClientConfiguration.kt
+++ b/core/src/main/java/com/schibsted/account/ClientConfiguration.kt
@@ -9,6 +9,7 @@ import android.os.Parcelable
 import android.support.annotation.VisibleForTesting
 import com.schibsted.account.common.util.Logger
 import com.schibsted.account.network.Environment
+import com.schibsted.account.network.ServiceHolder
 import com.schibsted.account.util.ConfigurationUtils
 
 /**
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/schibsted/account/ClientConfiguration.kt
```diff
@@ -61,6 +65,10 @@ data class ClientConfiguration(
         @JvmStatic
         fun set(clientConfiguration: ClientConfiguration) {
             currentConfig = clientConfiguration
+
+            ServiceHolder.oAuthService = OAuthService(get().environment, ServiceHolder.defaultClient)
```
zamzterz: This would be better encapsulated in a method in `ServiceHolder`:
```kotlin
internal fun reset() {
  oAuthService = OAuthService(ClientConfiguration.get().environment, defaultClient)
  clientService = ClientService(ClientConfiguration.get().environment, defaultClient)
  passwordlessService = PasswordlessService(ClientConfiguration.get().environment, defaultClient)
}
```

which would be called here as `ServiceHolder.reset()`.

JakubMosakowski: Yes, I agree!

Fixed.

zamzterz:
Superseded by #397 (for Travis sake).

PR #398, opened 2019-08-09
Fix pre-lollipop vector drawables crash

Swirastlynn:
According to current appcompat libraries setup. Tested on API 19 and 28.

There are different approaches to solution:
https://stackoverflow.com/questions/36867298/using-android-vector-drawables-on-pre-lollipop-crash
https://stackoverflow.com/questions/34417843/how-to-use-vector-drawables-in-android-api-lower-21/40523623#40523623

But for our current compat libraries setup this is the one which works.


zamzterz:
Superseded by #399 for Travis sake.

PR #402, opened 2019-08-22
Added missing callback invocation on success.

JakubMosakowski:
If applied, this commit will fix the issue of using onSuccessFun with logout.

For now, it is a bit misleading that we can pass onSuccessFun to logout method but it will never be invoked.


sa-mao:
Will be close as #403 supersedes it. Much thanks for helping with this one :)

PR #406, opened 2019-08-29
Focus on the input field and pop up keyboard when opening activity

Nilzor:
Focuses the input field and pops up the soft keyboard when the user enteres one of these three views:
- Username
- Password
- SMS (PhoneInputView)

Exception: if the user prefills the email field through the use of Google Smartlock.

In addition it makes sure we close the keyboard if the users minimizes the app while in input mode.

The solution is tested for all three views on two different phones:
- Samsung S8
- Huawei P20 Pro

PR replaces #366

### DIFF
--- a/ui/src/main/java/com/schibsted/account/ui/login/screen/identification/ui/EmailIdentificationFragment.kt
+++ b/ui/src/main/java/com/schibsted/account/ui/login/screen/identification/ui/EmailIdentificationFragment.kt
@@ -49,7 +49,7 @@ class EmailIdentificationFragment : AbstractIdentificationFragment(), Identifica
         super.onViewCreated(view, savedInstanceState)
         inputFieldView.inputField.requestFocus()
         if (inputFieldView.inputField.text.isNullOrBlank()) {
-            Logger.info(TAG, "Showing keyboard")
+            Logger.debug(TAG, "Showing keyboard")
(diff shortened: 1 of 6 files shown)


ui/src/main/java/com/schibsted/account/ui/login/screen/identification/ui/EmailIdentificationFragment.kt
```diff
@@ -44,6 +45,15 @@ class EmailIdentificationFragment : AbstractIdentificationFragment(), Identifica
         return view
     }
 
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        inputFieldView.inputField.requestFocus()
+        if (inputFieldView.inputField.text.isNullOrBlank()) {
+            Logger.info(TAG, "Showing keyboard")
```
zamzterz: I think this should be a debug log (instead of info).

Nilzor: Ok, fixed in e3a76b1

zamzterz:
Thanks for fixing it so quickly! Closing it in favor of #407 just for Travis.

Issue #405, opened 2019-08-29
Dismissing the SmartLock dialog causes reload of the username Activity


Nilzor:
To reproduce:
- Open the solution and modify the example app to [ENABLE smart lock](https://github.com/schibsted/account-sdk-android/blob/master/example/src/main/java/com/schibsted/account/example/MainActivity.java#L115)
- Open the app, click login and click _outside_ of the popup dialog from SmartLock
- Notice how the acitvity restarts

zamzterz:
This is intended in the example app implementation; the relevant parts is that 1) the SDK treating the [dismissal of SmartLock as a failure], and 2) the example app [restarting the login with SmartLock disabled](https://github.com/schibsted/account-sdk-android/blob/880de7adf6a0ed0849870f517cdc5ac171af9783/example/src/main/java/com/schibsted/account/example/MainActivity.java#L154).

I agree it's a bit unnecessary to restart the Activity. Maybe we should treat dismissal of the dialog or any other failure related to SmartLock as a failure only if `SmartLockMode.FORCED` is used.

Nilzor:
Sounds like a plan.

Note that it's not only the example app that is affected, but also production apps using the `ui` module

zamzterz:
@Nilzor This has been fixed in #408, if you're able to help out with review/testing it would be great. 🙂

PR #408, opened 2019-09-03
Ignore SmartLock failure

zamzterz:
To avoid unnecessarily requiring the login activity to be restarted if
SmartLock failed or was cancelled.

### Testing instructions
See #405.

### DIFF
--- a/smartlock/README.md
+++ b/smartlock/README.md
@@ -4,14 +4,14 @@ The smartlock module can be added to the UIs with the following dependency. The 
 implementation "com.schibsted.account:account-sdk-android-smartlock:<VERSION>"
 ```
 
-There are three smartlock modes:
-- SmartlockMode.DISABLED: The SDK will not attempt to log the user in with smartlock. 
-- SmartlockMode.ENABLED: The SDK will attempt to log the user in with smartlock. If it fails but smartlock managed to get the user identifier, the usual login flow will be launched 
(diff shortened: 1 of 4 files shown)


Nilzor:
Looks good to me, tested successfully on Huawei P20 Pro. 

But I'm not sure what "FORCED" is supposed to do? What's the difference between Forced and Enabled smart lock mode? They behave equally in the example app

zamzterz:
Previously the only difference was that when using `SmartlockMode.FORCED`, not having any stored credentials was treated as a failure.

Now there's the additional difference on what is treated as a failure. Per the (updated) documentation:
> In any case of failure when using the SmartlockMode.FORCED mode you will be notified in `onActivityResult` with the result code `AccountUi.SMARTLOCK_FAILED`.

Meaning if you dismiss the SmartLock dialog or there are no credentials stored when using `SmartlockMode.FORCED`, that is when you'll see the activity restart in the example app (due to [`AccountUi.SMARTLOCK_FAILED` being sent back to the starting activity](https://github.com/schibsted/account-sdk-android/blob/c6892cc4d119601442eb7b5299679f38ce7f1dca/example/src/main/java/com/schibsted/account/example/MainActivity.java#L154)).
But if you're using `SmartlockMode.ENABLED` the user is allowed to dismiss the dialog and no stored credentials are not treated as a failure (the user can simply enter their credentials).

PR #412, opened 2019-10-15
Pre-fill email if possible from previous login.

zamzterz:


ui/src/main/java/com/schibsted/account/ui/login/screen/identification/ui/EmailIdentificationFragment.kt
```diff
@@ -57,7 +58,14 @@ class EmailIdentificationFragment : AbstractIdentificationFragment(), Identifica
     public override fun prefillIdentifier(identifier: String?) {
         Logger.info(TAG, "Attempting to prefill  email")
         if (identifier.isNullOrEmpty()) {
-            Logger.info(TAG, "email wasn't found")
+            Logger.info(TAG, "email wasn't found in config")
+            val storedEmailPrefillValue: String? = this.context?.let { KeyValueStore(it).readEmailPrefillValue() }
+            storedEmailPrefillValue?.let {
+                if (EmailValidationRule.isValid(it)) {
```
sa-mao: Not sure if we should validate the email at this point.

zamzterz: You're right, if it's stored it should be valid. 👍

PR #433, opened 2020-01-23
Fix tests in common

birukoff:
ObservableField.Single behaves really weird :)
It looks like it is by design but I suspect this kind of design can be a source of bugs.

### DIFF
--- a/common/src/main/java/com/schibsted/account/common/lib/ObservableField.kt
+++ b/common/src/main/java/com/schibsted/account/common/lib/ObservableField.kt
@@ -6,15 +6,37 @@ package com.schibsted.account.common.lib
 
 import kotlin.properties.Delegates
 
-open class ObservableField<T>(initialValue: T) {
+/**
+ * Allows to reactively observe value and its changes. Emitted values are guaranteed
+ * to be distinct.
+ */
+class ObservableField<T>(initialValue: T) {
+
+    /**
+     * Interface for value observers.
+     */
     interface Observer<T> {
+
+        /**
+         * This method is called when value changes. New value is guaranteed to be not equal
+         * to the old value.
+         */
         fun onChange(newValue: T)
     }
 
+    /**
+     * Interface for value observers that are supposed to be invoked only once after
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/schibsted/account/common/lib/ObservableField.kt
```diff
@@ -6,15 +6,37 @@ package com.schibsted.account.common.lib
 
 import kotlin.properties.Delegates
 
-open class ObservableField<T>(initialValue: T) {
```
birukoff: No longer `open`

common/src/main/java/com/schibsted/account/common/lib/ObservableField.kt
```diff
@@ -6,15 +6,37 @@ package com.schibsted.account.common.lib
...
         fun onChange(newValue: T)
     }
 
+    /**
+     * Interface for value observers that are supposed to be invoked only once after
+     * a value change. Note, if an Single is added to ObservableField with flag
+     * `notifyInitially = true`, it will be called immediately AND it will be called again
+     * after a value change.
+     */
     interface Single<T> : Observer<T>
```
birukoff: Called twice in some conditions - WTF?

common/src/main/java/com/schibsted/account/common/util/Util.kt
```diff
@@ -10,18 +10,32 @@ import android.util.Base64
...
+ * Returns string value to be used for basic auth header ("Basic " + base64-encoded string
+ * that contains clientId and clientSecret separated by a colon).
+ */
 fun createBasicAuthHeader(clientId: String, clientSecret: String): String =
         "Basic ${encodeBase64("$clientId:$clientSecret")}"
 
+/**
+ * Encodes given string as Base64.
+ */
 fun encodeBase64(str: String): String = Base64.encodeToString(str.toByteArray(Charsets.UTF_8), Base64.NO_WRAP)
```
birukoff: Not unit-testable, because relies on OS.

PR #434, opened 2020-01-24
Minor fixes to ConfigurationUtils and its tests

birukoff:


core/src/main/java/com/schibsted/account/ClientConfiguration.kt
```diff
@@ -67,11 +67,11 @@ data class ClientConfiguration(
         }
 
         @VisibleForTesting(otherwise = VisibleForTesting.PROTECTED)
-        internal fun fromParams(params: Map<String, Any>): ClientConfiguration {
```
birukoff: It is always `Map<String, String>`, not `Map<String, Any>`

core/src/main/java/com/schibsted/account/util/ConfigurationUtils.kt
```diff
@@ -21,19 +30,19 @@ object ConfigurationUtils {
 
     @VisibleForTesting(otherwise = VisibleForTesting.PACKAGE_PRIVATE)
     internal fun getConfigResourceStream(path: String): InputStream {
-        val stream = ConfigurationUtils::class.java.classLoader.getResourceAsStream(path)
-        require(stream != null) { "Missing configuration asset: $path" }
-        return stream
+        val stream = ConfigurationUtils::class.java.classLoader?.getResourceAsStream(path)
+        return checkNotNull(stream) { "Missing configuration asset: $path" }
```
birukoff: `require` is used for arguments when you want to make sure a caller is doesn't want us to do something insane.
`check` is used for returned values to signal that something goes unexpectedly wrong while doing some legit work.

core/src/test/java/com/schibsted/account/util/ConfigurationUtilsTest.kt
```diff
@@ -41,7 +41,7 @@ class ConfigurationUtilsTest : WordSpec({
         }
 
         "trim blank space and lines" {
-            val lines = listOf("key1: value1", "key2:value2", " key3 :   value3  ", " ")
+            val lines = listOf("key1: value1", "key2:value2", " key3 :   value3  ", " ", " #comment")
```
birukoff: This case was not checked (and was broken)

core/src/main/java/com/schibsted/account/util/ConfigurationUtils.kt
```diff
@@ -21,19 +30,19 @@ object ConfigurationUtils {
...
     }
 
     @VisibleForTesting(otherwise = VisibleForTesting.PACKAGE_PRIVATE)
-    internal fun parseConfigFile(lines: List<String>): Map<String, Any> {
+    internal fun parseConfigFile(lines: List<String>): Map<String, String> {
         return lines
-                .filter { it.trim().isNotEmpty() }
-                .filterNot { it.startsWith('#') }
+                .map { it.trim() }
```
birukoff: `trim` first, otherwise ` #comment` will go through

core/src/main/java/com/schibsted/account/util/ConfigurationUtils.kt
zamzterz: One could argue though that supplying an invalid path is a caller trying to do something crazy. And using `require` seems to make more sense to me as it results in `IllegalArgumentException` (instead of `IllegalStateException` with `check`).

birukoff: IMO, path is actually hard-coded, and since method is internal, we don't expect it to be called with any other arguments. The call might fail not because of incorrect argument but because the file itself is not present in the assets => illegal state.

PR #436, opened 2020-01-28
Minor fixes to ClientConfiguration.

birukoff:
Honestly, I think `ClientConfiguration` should be merged with `utils.ConfigurationUtils`.

### DIFF
--- a/core/src/main/java/com/schibsted/account/ClientConfiguration.kt
+++ b/core/src/main/java/com/schibsted/account/ClientConfiguration.kt
@@ -4,13 +4,13 @@
 
 package com.schibsted.account
 
-import android.os.Parcel
 import android.os.Parcelable
 import android.support.annotation.VisibleForTesting
 import com.schibsted.account.common.util.Logger
 import com.schibsted.account.network.Environment
 import com.schibsted.account.network.ServiceHolder
 import com.schibsted.account.util.ConfigurationUtils
+import kotlinx.android.parcel.Parcelize
 
 /**
  * This contains the client configuration as found in the self service pages in Schibsted account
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/schibsted/account/ClientConfiguration.kt
```diff
@@ -27,9 +27,9 @@ data class ClientConfiguration(
 ) : Parcelable {
 
     constructor(source: Parcel) : this(
-            source.readString(),
-            source.readString(),
-            source.readString()
+            source.readString()!!,
+            source.readString()!!,
+            source.readString()!!
```
birukoff: Nullability warnings here

core/src/main/java/com/schibsted/account/ClientConfiguration.kt
```diff
@@ -45,18 +45,15 @@ data class ClientConfiguration(
         private const val KEY_ID = "clientId"
         private const val KEY_SECRET = "clientSecret"
         private const val KEY_ENVIRONMENT = "environment"
-        private var currentConfig: ClientConfiguration? = null
+        private lateinit var currentConfig: ClientConfiguration
 
         @JvmStatic
         fun get(): ClientConfiguration {
-            val conf = currentConfig
```
birukoff: I see no reason for this internal var

core/src/main/java/com/schibsted/account/ClientConfiguration.kt
```diff
@@ -45,18 +45,15 @@ data class ClientConfiguration(
         private const val KEY_ID = "clientId"
         private const val KEY_SECRET = "clientSecret"
         private const val KEY_ENVIRONMENT = "environment"
-        private var currentConfig: ClientConfiguration? = null
+        private lateinit var currentConfig: ClientConfiguration
```
birukoff: It should be `lateinit... ClientConfiguration` instead of `ClientConfiguration?`

core/src/main/java/com/schibsted/account/ClientConfiguration.kt
```diff
@@ -20,43 +20,27 @@ import com.schibsted.account.util.ConfigurationUtils
  * @see <a href="https://selfservice.login.schibsted.com/login">Self-Service PRO</a>
  * @see <a href="https://selfservice.identity-pre.schibsted.com">Self-Service PRE</a>
  */
+@Parcelize
```
birukoff: `@Parcelize` eliminates the boilerplate of `Parcelable`

core/src/main/java/com/schibsted/account/ClientConfiguration.kt
```diff
@@ -20,43 +20,27 @@ import com.schibsted.account.util.ConfigurationUtils
...
+        private lateinit var currentConfig: ClientConfiguration
 
         @JvmStatic
         fun get(): ClientConfiguration {
-            val conf = currentConfig
-            if (conf == null) {
-                val newConf = fromParams(ConfigurationUtils.paramsFromAssets())
-                currentConfig = newConf
-                return newConf
+            if (!::currentConfig.isInitialized) {
```
birukoff: This code is not thread-safe. Should it be thread-safe?

zamzterz: I don't think it needs to be - was it thread-safe before this change?

zamzterz:
Also, I agree that merging `ClientConfiguration` with `utils.ConfigurationUtils` would be a good idea (in a separate PR). 👍

core/src/main/java/com/schibsted/account/ClientConfiguration.kt
birukoff: No, it wasn't. It just looks suspiciously similar to double-checking lazy initialization which is usually used in multi-threaded context.

zamzterz: Then no change, so should be safe (or at least no more unsafe than it was before 🙂).

PR #440, opened 2020-02-18
Force RSA key refresh on upgrade to v1.11.2

birukoff:
The branch is just for releasing a hotfix for 1.11.2. There are no plans to merge it to master.

### DIFF
--- a/core/src/main/java/com/schibsted/account/persistence/SessionStorageDelegate.kt
+++ b/core/src/main/java/com/schibsted/account/persistence/SessionStorageDelegate.kt
@@ -39,6 +39,8 @@ internal class SessionStorageDelegate(
     }
 
     private fun writeDataToPersistence(data: List<UserPersistence.Session>, prefs: SharedPreferences) {
+        refreshUncertainKeys(prefs)
+
         val dataToPersist = GSON.toJson(data)
         var aesKey: SecretKey? = null
 
@@ -105,6 +107,7 @@ internal class SessionStorageDelegate(
             }
 
             return sessions?.also {
+                refreshUncertainKeysAndPersist(prefs, it)
                 updateKeyExpiry(prefs, it)
             }
         } else {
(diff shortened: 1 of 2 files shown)


birukoff:
Again, a lot of whitespace changes. The actual change is fairly small.

core/src/main/java/com/schibsted/account/persistence/SessionStorageDelegate.kt
```diff
@@ -1,231 +1,270 @@
...
+        set(value) {
+            appContext.getSharedPreferences(preferenceFilename, Context.MODE_PRIVATE).edit()
+                    .run {
+                        putBoolean(KEYS_REFRESHED, value)
+                        apply()
+                    }
+        }
+
+    /**
+     * Workaround for issue, introduced in v1.5.0
```
zamzterz: Would be nice to describe the actual issue, the comment as-is is a bit non-descriptive (=will be hard to remember what this references in 6 months).

birukoff:
Generally, it will be described in a separate PR to master. This branch
will not be merged, it is only to release 1.11.2 (because master is way
ahead of it).

release.sh
```diff
@@ -7,7 +7,6 @@ if [ "$TRAVIS_PULL_REQUEST" = "false" ] && [ ! -z "$TRAVIS_TAG" ]; then
     ./gradlew :core:bintrayUpload
     ./gradlew :ui:bintrayUpload
     ./gradlew :smartlock:bintrayUpload
-    ./deploy_docs.sh
```
sa-mao: Why remove this line?

core/src/main/java/com/schibsted/account/persistence/SessionStorageDelegate.kt
```diff
@@ -1,231 +1,270 @@
...
+        isKeyExpirationCertain = true
+    }
+
+    /**
+     * Workaround for issue, introduced in v1.5.0
+     */
+    private fun refreshUncertainKeysAndPersist(prefs: SharedPreferences,
+                                               sessions: List<UserPersistence.Session>) {
+        if (isKeyExpirationCertain) return
+
```
sa-mao: Not urgent but maybe add some logging to indicate the refresh was triggered because of uncertainty.

release.sh
birukoff: It will regenerate javadoc. The current javadoc is already at 1.12. We don't want to reset it back.

core/src/main/java/com/schibsted/account/persistence/SessionStorageDelegate.kt
birukoff: I will add but don't expect much of it - it is a one time thing

PR #443, opened 2020-02-24
Force RSA key refresh on upgrade to 1.12

birukoff:
Just like we did for 1.11.2 but simpler.

### DIFF
--- a/core/src/androidTest/java/com/schibsted/account/persistence/EncryptionKeyProviderTest.kt
+++ b/core/src/androidTest/java/com/schibsted/account/persistence/EncryptionKeyProviderTest.kt
@@ -229,15 +229,26 @@ class EncryptionKeyProviderTest {
         val provider = EncryptionKeyProvider.create(appContext)
         provider.refreshKeyPair()
 
-        val expiration = System.currentTimeMillis() + TimeUnit.DAYS.toMillis(90)
+        // 90 days is exactly the threshold when we consider the key too close to expiration:
+        val expiration = System.currentTimeMillis() + TimeUnit.DAYS.toMillis(90) - 100L
         prefs.edit().putLong(KEY_EXPIRATION, expiration).commit()
+
+        assertTrue(provider.isKeyCloseToExpiration())
+    }
+
+    @Test
+    fun keyIsCloseToExpiration_whenExpirationDateUnknown() {
+        val provider = EncryptionKeyProvider.create(appContext)
+        provider.refreshKeyPair()
+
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/schibsted/account/persistence/EncryptionKeyProvider.kt
```diff
@@ -90,7 +90,8 @@ private const val KEYSTORE_PROVIDER = "AndroidKeyStore"
 private const val PREFS_FILENAME = "IDENTITY_KEYSTORE"
 private const val PREFS_PRIVATE_KEY = "IDENTITY_PR_KEY_PAIR"
 private const val PREFS_PUBLIC_KEY = "IDENTITY_PU_KEY_PAIR"
-private const val PREFS_EXPIRATION = "KEY_PAIR_VALID_UNTIL"
+private const val PREFS_EXPIRATION = "IDENTITY_KEY_EXPIRATION_DATE"
```
birukoff: Changed key. Now all queries will return default value again

core/src/main/java/com/schibsted/account/persistence/EncryptionKeyProvider.kt
```diff
@@ -102,7 +103,7 @@ private abstract class BaseProvider(
     }
 
     protected var expiration: Long
-        get() = prefs.getLong(PREFS_EXPIRATION, NEVER)
+        get() = prefs.getLong(PREFS_EXPIRATION, UNKNOWN)
```
birukoff: Default value is "UNKNOWN" and all calls to `isKeyCloseToExpiration` will return true -> SessionStorageDelegate will force RSA key refresh.

core/src/androidTest/java/com/schibsted/account/persistence/EncryptionKeyProviderTest.kt
```diff
@@ -229,15 +229,29 @@ class EncryptionKeyProviderTest {
         val provider = EncryptionKeyProvider.create(appContext)
         provider.refreshKeyPair()
 
+        // 90 days is exactly the threshold when we consider the key too close to expiration:
         val expiration = System.currentTimeMillis() + TimeUnit.DAYS.toMillis(90)
         prefs.edit().putLong(KEY_EXPIRATION, expiration).commit()
 
+        // Wait until the key becomes too close to expiration:
+        Thread.sleep(100L)
```
zamzterz: Why not create the key with slightly shorter expiration instead of using `sleep`?

PR #444, opened 2020-02-25
Web flow login

zamzterz:
To pick up existing user session in the device browser, or for a completely new login via web flows (instead of native UI).


core/src/main/java/com/schibsted/account/util/DeepLink.kt
```diff
@@ -55,6 +56,67 @@ sealed class DeepLink {
         }
     }
 
+    class WebFlowLogin private constructor(code: String, isPersistable: Boolean, val codeVerifier: String) : ValidateAccount(code, isPersistable, arrayOf(OIDCScope.SCOPE_OPENID)) {
+        companion object {
+            private const val PARAM_STATE = "state"
+            private const val PREFERENCE_FILENAME = "WEB_FLOW_LOGIN"
```
birukoff: Why do you use a separate SharedPrefs file everywhere? Generally speaking, it is a bad practice since it causes very long IO the first time you access this particular SharedPrefs on this particular context. If done on the main thread, it will cause dropped frames and stuttering in animations.
Not saying that you should change it right here (if this pattern is used everywhere, probably, this is not a right PR for this) but this is a thing to keep in mind and to fix in future refactoring.

core/src/main/java/com/schibsted/account/util/DeepLink.kt
```diff
@@ -55,6 +56,67 @@ sealed class DeepLink {
...
+            private const val SHARED_PREFERENCES_OAUTH_STATE = "OAUTH_STATE"
+            private const val SHARED_PREFERENCES_CODE_VERIFIER = "CODE_VERIFIER"
+            private const val SHARED_PREFERENCES_PERSIST_USER = "PERSIST_USER"
+
+            data class Data(val oauthState: String, val codeVerifier: String, val persistUser: Boolean)
+
+            operator fun invoke(context: Context, uri: URI): WebFlowLogin? {
+                val prefs = context.applicationContext.getSharedPreferences(PREFERENCE_FILENAME, Context.MODE_PRIVATE)
```
birukoff: Java style :)
Kotlin style is using properties.

core/src/main/java/com/schibsted/account/util/DeepLink.kt
```diff
@@ -55,6 +56,67 @@ sealed class DeepLink {
...
+                    Logger.info(TAG, "WebFlowLogin: state is missing from response")
+                    return null
+                } else if (state != storedData.oauthState) {
+                    Logger.info(TAG, "WebFlowLogin: unexpected state")
+                    return null
+                }
+
+                clearData(prefs)
+                val code = uri.getQueryParam(PARAM_CODE)
+                if (code == null) {
```
birukoff: What about this:
`val code = uri.getQueryParam(PARAM_CODE) ?: return null.also {
    Logger.info(TAG, "WebFlowLogin: code is missing from response")
}`

core/src/main/java/com/schibsted/account/util/DeepLink.kt
```diff
@@ -55,6 +56,67 @@ sealed class DeepLink {
...
+                    putBoolean(SHARED_PREFERENCES_PERSIST_USER, data.persistUser)
+                    apply()
+                }
+            }
+
+            private fun getData(prefs: SharedPreferences): Data {
+                val oauthState = prefs.getString(SHARED_PREFERENCES_OAUTH_STATE, "")
+                val codeVerifier = prefs.getString(SHARED_PREFERENCES_CODE_VERIFIER, "")
+                val persistUser = prefs.getBoolean(SHARED_PREFERENCES_PERSIST_USER, true)
+                return Data(oauthState!!, codeVerifier!!, persistUser)
```
birukoff: `!!` is VERY, VERY bad! Big no-no!

core/src/main/java/com/schibsted/account/util/DeepLink.kt
```diff
@@ -55,6 +56,67 @@ sealed class DeepLink {
...
+            private const val SHARED_PREFERENCES_PERSIST_USER = "PERSIST_USER"
+
+            data class Data(val oauthState: String, val codeVerifier: String, val persistUser: Boolean)
+
+            operator fun invoke(context: Context, uri: URI): WebFlowLogin? {
+                val prefs = context.applicationContext.getSharedPreferences(PREFERENCE_FILENAME, Context.MODE_PRIVATE)
+                val storedData = getData(prefs)
+                val state = uri.getQueryParam(PARAM_STATE)
+
+                if (state == null) {
```
birukoff: I think `when` would look way better then `if` here

core/src/main/java/com/schibsted/account/util/DeepLink.kt
```diff
@@ -55,6 +56,67 @@ sealed class DeepLink {
...
+                with(prefs.edit()) {
+                    putString(SHARED_PREFERENCES_OAUTH_STATE, data.oauthState)
+                    putString(SHARED_PREFERENCES_CODE_VERIFIER, data.codeVerifier)
+                    putBoolean(SHARED_PREFERENCES_PERSIST_USER, data.persistUser)
+                    apply()
+                }
+            }
+
+            private fun getData(prefs: SharedPreferences): Data {
+                val oauthState = prefs.getString(SHARED_PREFERENCES_OAUTH_STATE, "")
```
birukoff: `val oauthState = prefs.getString(SHARED_PREFERENCES_OAUTH_STATE, null) ?: ""` - will make `oauthState` non-null `String`

core/src/main/java/com/schibsted/account/util/DeepLink.kt
birukoff: Example: https://github.com/schibsted/account-sdk-android/blob/master/core/src/main/java/com/schibsted/account/persistence/EncryptionKeyProvider.kt#L101

core/src/main/java/com/schibsted/account/util/DeepLink.kt
zamzterz: Not sure what the original reason was, but as you're hinting at I kept it like this for consistency.

core/src/main/java/com/schibsted/account/util/DeepLink.kt
zamzterz: I saw that as an option, but to me that `return null.also` reads strangely 🙂 
But could be my Kotlin-inexperience, so if you think that's better I can change it.

core/src/main/java/com/schibsted/account/util/DeepLink.kt
zamzterz: Fixed 👍

birukoff:
There are failing tests though

PR #451, opened 2020-03-02
Refactor UserPersistence.

birukoff:
I had to extract `ResumeDelegate` and leave it without tests because (1) right now this piece of code is untestable, and (2) making it testable would require refactoring of `User` and `Agreements` which are beyond the scope of this PR. It is rather small, so in my opinion it should be safe (for now).

`AgreementStorage` deals with storing values, no business logic.
`AgreementCache` deals with business logic: checks if cached agreement is still valid, checks if user ID matches, etc.


core/src/main/java/com/schibsted/account/persistence/AgreementStorage.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    companion object {
+        private const val TAG = "AgreementCache"
+        private const val PREFERENCE_FILENAME = "IDENTITY_PREFERENCES"
+        private const val KEY_AGR_CACHE = "AGR_CACHE"
+        private const val DELIMITER = "|"
+        private const val DATE_PARSING_FORMAT = "yyyy-MM-dd'T'HH:mm:ss"
+
+        private val sdf = object : ThreadLocal<SimpleDateFormat>() {
+            override fun initialValue() = SimpleDateFormat(DATE_PARSING_FORMAT, Locale.US)
```
birukoff: There is bug here that I didn't fix: date is stored without timezone. When user moves between timezones, it will affect cache expiration. 
Since we want to have random cache expiration anyway, I don't consider this a big problem.
At the same time, fixing it would (1) introduce extra code that will not improve readability and (2) potentially create other bugs.
So, I left this bug as it was.

core/src/main/java/com/schibsted/account/persistence/AgreementCache.kt
```diff
@@ -0,0 +1,45 @@
+package com.schibsted.account.persistence
+
+import java.util.*
+import java.util.concurrent.TimeUnit
+
+/**
+ * Stores user ID with some random future expiration Date. If queried with the same user ID
```
zamzterz: Very descriptive summary - nice! 👍

PR #452, opened 2020-03-02
Use publication's artifactId for POM generation of ProjectDependency.

birukoff:
Publication's artifact ID is set by each project separately, and it is not equal to project's name (for example, `common`'s name is "common" but artifactId is "account-sdk-android-common").
When generating POM with dependencies, we need to use artifactId instead of project's name.

### DIFF
--- a/buildSrc/src/main/java/PomHelpers.kt
+++ b/buildSrc/src/main/java/PomHelpers.kt
@@ -1,9 +1,9 @@
 import groovy.util.Node
 import org.gradle.api.Project
 import org.gradle.api.XmlProvider
-import org.gradle.api.artifacts.Dependency
-import org.gradle.api.artifacts.ExcludeRule
-import org.gradle.api.artifacts.ModuleDependency
+import org.gradle.api.artifacts.*
+import org.gradle.api.plugins.ExtensionAware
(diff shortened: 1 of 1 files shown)


birukoff:
@zamzterz I recall we talked about whether to use or not to use third-party Gradle plugins to generate POM. This case wouldn't be handled by that plugin, and we would be screwed.
This is nice illustration how having them becomes more expensive then writing simple stuff on our own.

buildSrc/src/main/java/PomHelpers.kt
```diff
@@ -50,8 +50,8 @@ fun MavenPom.fillGenericDetails(project: Project) {
 fun XmlProvider.collectDependencies(project: Project) {
 
     fun Node.appendExclusion(excludeRule: ExcludeRule) = appendNode("exclusion").apply {
-        appendNode("groupId", excludeRule.group ?: "*")
-        appendNode("artifactId", excludeRule.module ?: "*")
+        appendNode("groupId", excludeRule.group as? String ?: "*")
+        appendNode("artifactId", excludeRule.module as? String ?: "*")
```
birukoff: There was a warning here that "excludeRule.group is always not null". In fact, we know for sure that it is nullable. So, here is the fix.

PR #459, opened 2020-03-12
Don't depend on user_id in token response.

zamzterz:
It's not defined in OAuth/OIDC so not supported by oauth-service.

### DIFF
--- a/core/src/main/java/com/schibsted/account/persistence/ResumeDelegate.kt
+++ b/core/src/main/java/com/schibsted/account/persistence/ResumeDelegate.kt
@@ -19,16 +19,26 @@ internal class ResumeDelegate(private val agreementCache: AgreementCache) {
     fun proceed(token: UserToken,
                 success: (user: User) -> Unit,
                 failure: (error: ClientError) -> Unit) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/schibsted/account/persistence/ResumeDelegate.kt
```diff
@@ -19,16 +19,26 @@ internal class ResumeDelegate(private val agreementCache: AgreementCache) {
     fun proceed(token: UserToken,
                 success: (user: User) -> Unit,
                 failure: (error: ClientError) -> Unit) {
-        val user = User(token, isPersistable = true)
-        if (agreementCache.hasValidAgreement(user.userId.id)) {
-            success(user)
+        val user = try {
+             User(token, isPersistable = true)
+        } catch (e: IllegalArgumentException) {
+            null
```
birukoff: This makes my eyes bleed!

birukoff:
Can you explain me why is this change needed?

zamzterz:
It is needed as we'll replace the backend service for this request. And the new service doesn't include `user_id` in the response so it should be read from the ID Token instead.

Issue #457, opened 2020-03-05
Security alert: App contains unsafe cryptographic encryption pattern


damianpetla:
### Summary
According to Google Play console this class `com.schibsted.account.persistence.PersistenceEncryption.aesEncrypt` uses unsafe pattern and Google see that as security vulnerability and suggest following this help site https://support.google.com/faqs/answer/9450925

### Details
SDK 1.9.0

PR #460, opened 2020-03-20
Use Random IV in EncryptionUtils.

birukoff:
So, this is supposed to fix https://github.com/schibsted/account-sdk-android/issues/457

1) `EncryptionUtils` contains the fix for static IV. The old code remains there so that we can migrate existing encrypted data.
2) `SessionStorageLegacy` uses old method to decrypt existing data.
3) `SessionStorageDelegate` uses new SharedPrefs location to store key/data because the old stuff is not readable using this new method, and we have to migrate it first. In order to migrate data, before accessing this new location it queries `SessionStorageLegacy` if it has any stored data. If it is there, it is re-saved using new encryption method and then deleted.


core/src/main/java/com/schibsted/account/persistence/EncryptionUtils.kt
```diff
@@ -38,20 +39,39 @@ internal interface EncryptionUtils {
...
     fun aesEncrypt(subjectToEncrypt: ByteArray, aesKey: SecretKey): ByteArray =
             with(Cipher.getInstance(AES_TRANSFORM)) {
-                init(Cipher.ENCRYPT_MODE, aesKey, IvParameterSpec(ByteArray(16)))
-                doFinal(subjectToEncrypt)
+                val vector = ByteArray(16).also {
+                    SecureRandom().nextBytes(it)
+                }
+                init(Cipher.ENCRYPT_MODE, aesKey, IvParameterSpec(vector))
+                val cipherText = doFinal(subjectToEncrypt) ?: return@with null
+                vector + cipherText
```
birukoff: This is standard way of storing IV. It doesn't have to be secret.

core/src/main/java/com/schibsted/account/persistence/SessionStorageDelegate.kt
```diff
@@ -24,8 +24,8 @@ internal class SessionStorageDelegate(
 
     companion object {
         private const val TAG = "SessionStorageDelegate"
-        private const val PREF_KEY_DATA = "IDENTITY_SESSIONS"
-        private const val PREF_KEY_AES = "IDENTITY_AES_PREF_KEY"
+        private const val PREF_KEY_DATA = "com.schibsted.account.persistence.sessions"
+        private const val PREF_KEY_AES = "com.schibsted.account.persistence.aeskey"
```
birukoff: These key are created with another future optimization in mind - eventually we should use only 1 SharedPrefs file for all our preferences.

core/src/main/java/com/schibsted/account/persistence/SessionStorageDelegate.kt
```diff
@@ -24,8 +24,8 @@ internal class SessionStorageDelegate(
 
     companion object {
         private const val TAG = "SessionStorageDelegate"
-        private const val PREF_KEY_DATA = "IDENTITY_SESSIONS"
-        private const val PREF_KEY_AES = "IDENTITY_AES_PREF_KEY"
+        private const val PREF_KEY_DATA = "com.schibsted.account.persistence.SessionStorageDelegate.sessions"
+        private const val PREF_KEY_AES = "com.schibsted.account.persistence.SessionStorageDelegate.aeskey"
```
birukoff: These key are created with another future optimization in mind - eventually we should use only one SharedPrefs file for all our preferences.

core/src/main/java/com/schibsted/account/persistence/SessionStorageLegacy.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    private fun fetchSessions(): List<UserPersistence.Session>? {
+        val encryptedKey = prefs.getBytes(KEY_AES) ?: return null
+        val privateRsaKey = encryptionKeyProvider.keyPair.private
+        val decryptedKey = encryptionUtils.rsaDecrypt(encryptedKey, privateRsaKey)
+        val secretKey = encryptionUtils.recreateAesKey(decryptedKey)
+
+        val encryptedData = prefs.getBytes(KEY_DATA) ?: return null
+        // Old SDK versions used to encrypt data using empty ByteArray:
+        val iv = ByteArray(16)
```
birukoff: This is bug 1 - using empty array as initialization vector

core/src/main/java/com/schibsted/account/persistence/SessionStorageLegacy.kt
```diff
@@ -0,0 +1,82 @@
...
+        val decryptedData = encryptionUtils.aesDecrypt(encryptedData, secretKey, iv)
+        // Old SDK versions used to Base64-encode data twice:
+        val decodedData = decryptedData.decodeBase64()
```
birukoff: This is bug 2 - data was encoded to Base64 twice, and had to decoded twice as well.

core/src/main/java/com/schibsted/account/persistence/SessionStorageDelegate.kt
```diff
@@ -24,8 +24,10 @@ internal class SessionStorageDelegate(
 
     companion object {
         private const val TAG = "SessionStorageDelegate"
-        private const val PREF_KEY_DATA = "IDENTITY_SESSIONS"
-        private const val PREF_KEY_AES = "IDENTITY_AES_PREF_KEY"
+        private const val PREFIX = "com.schibsted.account.persistence.SessionStorageDelegate"
+        private const val PREF_MIGRATED = "$PREFIX.migrationCompleted"
+        private const val PREF_KEY_DATA = "$PREFIX.sessions"
+        private const val PREF_KEY_AES = "$PREFIX.aeskey"
```
birukoff: These key are created with another future optimization in mind - eventually we should use only one SharedPrefs file for all our preferences.

core/src/main/java/com/schibsted/account/persistence/SessionStorageDelegate.kt
```diff
@@ -187,4 +191,34 @@ internal class SessionStorageDelegate(
...
+     * Sessions encrypted in the new way are stored in a new location in SharedPreferences.
+     * This lets us migrate the existing session list and not cause unintended "sign-outs".
+     */
+    private val legacy: SessionStorageLegacy? by lazy {
+        // SessionStorageLegacy uses SharedPrefs. Calling its methods will involve disk IO.
+        // Having this extra check allows us to avoid needless IO:
+        if (prefs.getBoolean(PREF_MIGRATED, false)) {
+            null
+        } else {
+            SessionStorageLegacy(appContext, encryptionKeyProvider, encryptionUtils)
```
birukoff: The code above ensures that `SessionStorageLegacy` is created only for the very first time in order to facilitate data migration. On later launches it will be null == no needless objects, no GC, no IO.

PR #467, opened 2020-06-12
Separate client and user token endpoints

birukoff:
The first step. The next step is separating network responses from internal data classes.
HttpLoggingInterceptor added temporarily and will be removed before releasing. It is just hard to understand things without it, and cumbersome to add-remove it for each branch.
Also, minor cleanup.


core/src/main/java/com/schibsted/account/model/TokenTypes.kt
```diff
@@ -4,7 +4,8 @@
...
 package com.schibsted.account.model
 
-import com.schibsted.account.network.response.TokenResponse
+import com.schibsted.account.network.response.ClientTokenResponse
+import com.schibsted.account.network.response.UserTokenResponse
 
-typealias ClientToken = TokenResponse
-typealias UserToken = TokenResponse
+typealias ClientToken = ClientTokenResponse
+typealias UserToken = UserTokenResponse
```
birukoff, line 11: These will become their own classes soon.

core/src/main/java/com/schibsted/account/network/response/ClientTokenResponse.kt
```diff
@@ -0,0 +1,61 @@
...
+
+data class ClientTokenResponse(
+    @SerializedName("id_token") val idToken: String? = null,
+    @SerializedName("user_id") val userId: String,
+    @SerializedName("access_token") val serializedAccessToken: String,
+    @SerializedName("refresh_token") val refreshToken: String?,
+    @SerializedName("scope") val scope: String,
+    @SerializedName("token_type") val tokenType: String,
+    @SerializedName("expires_in") val expiresIn: Int
+) : Parcelable {
```
birukoff, line 21: These `...TokenResponse` classes will cleaned up by applying `@Parcelize` annotation.

core/src/main/java/com/schibsted/account/network/service/authentication/OAuthService.kt
```diff
@@ -30,14 +31,14 @@ class OAuthService(@Environment environment: String, okHttpClient: OkHttpClient)
         redirectUri: String,
         scopes: Array<String>?,
```
birukoff, line 32: Shouldn't it be `vararg`, like everywhere else?

core/src/main/java/com/schibsted/account/network/service/authentication/OAuthService.kt
```diff
@@ -100,7 +101,7 @@ class OAuthService(@Environment environment: String, okHttpClient: OkHttpClient)
      * @param refreshToken The refresh token to use during the refresh.
      * successful.
      */
-    fun refreshToken(clientId: String, clientSecret: String, refreshToken: String): Call<TokenResponse> {
+    fun refreshToken(clientId: String, clientSecret: String, refreshToken: String): Call<UserTokenResponse> {
```
birukoff, line 104: `ClientToken` does not require "refreshing", does it?

core/src/main/java/com/schibsted/account/network/service/passwordless/PasswordlessService.java
```diff
@@ -23,11 +23,12 @@
 
 public class PasswordlessService extends BaseNetworkService {
 
-    public static final String PARAM_CONNECTION_EMAIL = "email";
-    public static final String PARAM_CONNECTION_SMS = "sms";
+    private static final String PARAM_CONNECTION_EMAIL = "email";
+    private static final String PARAM_CONNECTION_SMS = "sms";
```
birukoff, line 27: See no reason to keep public

core/src/main/java/com/schibsted/account/network/response/ClientTokenResponse.kt
```diff
@@ -0,0 +1,61 @@
...
+
+    fun bearerAuthHeader(): String = "Bearer $serializedAccessToken"
+
+    // This needs to be fail safe form Kotlin's null checks, as parsing done in Java can disregard null checks
```
zamzterz, line 25: Typo? form -> from

core/src/main/java/com/schibsted/account/network/service/authentication/OAuthService.kt
zamzterz: Probably - as far as I can think there's nothing that would warrant different type here.

core/src/main/java/com/schibsted/account/network/service/authentication/OAuthService.kt
```diff
@@ -30,14 +31,14 @@ class OAuthService(@Environment environment: String, okHttpClient: OkHttpClient)
...
-    ): Call<TokenResponse> {
+    ): Call<UserTokenResponse> {
 
         val params = mutableMapOf(
                 PARAM_GRANT_TYPE to GRANT_TYPE_AUTHORIZATION_CODE,
                 PARAM_CODE to authCode,
                 PARAM_REDIRECT_URI_UNDERSCORE to redirectUri)
-        scopes?.let { params.put(PARAM_SCOPE, scopes.joinToString { " " }) }
-        codeVerifier?.let { params.put(PARAM_CODE_VERIFIER, codeVerifier) }
+        scopes?.let { params.put(PARAM_SCOPE, it.joinToString(" ")) }
```
zamzterz, line 40: 👍

core/src/main/java/com/schibsted/account/network/service/authentication/OAuthService.kt
zamzterz: Exactly!

core/src/main/java/com/schibsted/account/util/KeyValueStore.java
```diff
@@ -116,17 +116,17 @@ public TokenResponse readAccessTokenCompat(final String clientSecret) {
      * @return The stored client token, if any. <code>null</code> otherwise.
      */
     @Nullable
-    public TokenResponse readClientAccessToken() {
+    public UserTokenResponse readClientAccessToken() {
```
zamzterz, line 119: Judging by the name of the method: shouldn't this have return type `ClientTokenResponse` and `writeClientToken` should accept `ClientTokenResponse`?

core/src/main/java/com/schibsted/account/network/response/ClientTokenResponse.kt
birukoff: This is a copy of the original file. I will clean it later.

core/src/main/java/com/schibsted/account/util/KeyValueStore.java
birukoff: Good catch!

PR #469, opened 2020-06-19
Separate Client/UserToken from Client/UserTokenResponse

birukoff:
WIP


core/src/main/java/com/schibsted/account/session/User.kt
```diff
@@ -19,37 +18,47 @@ import com.schibsted.account.model.UserId
...
-    constructor(parcel: Parcel) : this(parcel.readParcelable<UserTokenResponse>(UserTokenResponse::class.java.classLoader),
-            parcel.readInt() != 0)
-
-    @Volatile
-    internal var token: UserToken? = token
-        private set
-
-    val userId: UserId = UserId.fromTokenResponse(token)
-
+//class User(userTokenResponse: UserTokenResponse, val isPersistable: Boolean) : Parcelable {
```
zamzterz, line 30: This should be removed?

core/src/main/java/com/schibsted/account/session/User.kt
```diff
@@ -123,15 +132,16 @@ class User(token: UserToken, val isPersistable: Boolean) : Parcelable {
                 ClientConfiguration.get().clientSecret, refreshToken).execute()
 
         return if (resp.isSuccessful) {
-            this.token = requireNotNull(resp.body(), { "Unable to parse token from successful response" })
+            val userTokenResponse = checkNotNull(resp.body(), { "Response is successful ut body is null" })
```
zamzterz, line 135: "ut"? 🙂

core/src/main/java/com/schibsted/account/engine/controller/PasswordlessController.kt
```diff
@@ -103,8 +100,10 @@ class PasswordlessController @JvmOverloads constructor(
         if (res == null) {
             VerificationCode.request(provider, identifier) { verificationCode, callback ->
                 VerifyCodeOperation(identifier, passwordlessToken, verificationCode, scopes, { callback.onError(it.toClientError()) },
-                        { token ->
-                            val user = User(token, verificationCode.keepLoggedIn)
+                        {
+                            val userToken = UserToken(it)
+                            val userId = UserId.fromUserTokenResponse(it)
```
zamzterz, line 105: Now this should accept `UserToken` param (e.g. since `UserTokenResponse` is no the type persisted in storage any more).

core/src/main/java/com/schibsted/account/network/response/UserTokenResponse.kt
```diff
@@ -20,42 +21,11 @@ data class UserTokenResponse(
     @SerializedName("expires_in") val expiresIn: Int
 ) : Parcelable {
 
-    fun bearerAuthHeader(): String = "Bearer $serializedAccessToken"
```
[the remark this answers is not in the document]
birukoff: `UserTokenResponse` is a data class that represents a _response from server_. `bearerAuthHeader` is an attribute that can be used when making _requests to server_. In other words, `UserTokenResponse` is data, auth header is how this data is used. It has no place in the data class. It may even not come from this data class (and if it comes, it is purely coincidental). This method violates single responsibility principle and was removed.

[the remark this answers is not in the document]
birukoff: I don't mind having a helper that produces auth token from response (and actually I have it in works). But I oppose having it here, where it doesn't belong.

PR #471, opened 2020-08-14
Migrate account-sdk-android to AndroidX

ziem:
I've migrated account-sdk-android to AndroidX. Unit and Android tests pass correctly. I also successfully logged in and registered a new account.

Screenshots:
Screenshot_1597404728

Closes #470

PR #472, opened 2020-08-17
Migrate to AndroidX

zamzterz:
Supersedes #471 just for Travis sake.


core/src/main/java/com/schibsted/account/AccountService.kt
```diff
@@ -66,7 +66,7 @@ class AccountService @JvmOverloads constructor(
 
     companion object {
         private const val TAG = "AccountService"
-        internal var localBroadcastManager: LocalBroadcastManager? = null
+        internal var localBroadcastManager: androidx.localbroadcastmanager.content.LocalBroadcastManager? = null
```
birukoff, line 69: why fully qualified path? It is imported above

ziem: This is fixed in 4bf88250d1d2a2710af5d0c91af6266dd962f1bc

zamzterz:
Closing this as the Travis setup has been fixed (#472) to allow tests running directly against #471.

PR #475, opened 2020-10-01
Upgrade OkHttp dependency

zamzterz:
For compatibility with e.g. latest Firebase Crashlytics.


birukoff (approved):
You need to update .travis.yaml too

birukoff:
Probably, this next release will require a big announcement, because this change makes it incompatible with SDK < 21

outlying (approved):

PR #477, opened 2021-02-04
Center sch logo with client logo

xserxses:
When client logo (on the left) is bigger (and can be up to 150dp) than Schibsted logo  - Schibsted logo (on right) gets misaligned - glued to the top.

BEFORE:
image

When this PR is merged it will be nicely centered.
AFTER:
image


zamzterz (approved):
Looks so much better 👍

PR #478, opened 2021-02-08
NestedScrollView must have only one child fix

ziem:
Hello!
We are a bit on a bleeding-edge when it comes to the AndroidX/Jetpack libraries. Recently, we upgraded the fragment library from 1.3.0-alpha06 to 1.3.0-rc02. Unfortunately, this new version is making account-sdk-android crash. We get `NestedScrollView` must have only one child when we click "Continue" after typing in our email. I found a solution that uses `androidx.fragment.app.FragmentContainerView`. I also had to update the fragment library from 1.1.0 to 1.2.0 to make it visible, but at least it isn't alpha or rc. :slightly_smiling_face:


zamzterz (approved):
LGTM 👍

